PR #250, opened 2014-09-22
use gill sans mt instead of just gill sans

hwbllmnn:
The gill sans font on windows is a very wide font and breaks most layouts.


hwbllmnn:
Probably not an issue any more with bootstrap.

PR #251, opened 2014-09-23
Extent fields

hwbllmnn:
Makes the extent values available for printing.

Adds fields to use in templates: left, top, right, bottom.

Includes #250


hwbllmnn:
Print was refactored since.

PR #274, opened 2015-04-26
- core ru translation

mserge:
I translated corebundle translation for Russian language


astroidex:
Hello mserge, thanks for the translations.

PR #293, opened 2015-07-10
Hotfix/wms ratio buffer

psch:


hwbllmnn:
We should really think if we want this in a point release - it requires changes to the db structure.

hwbllmnn:
Ok, my bad, it contains comments that just look like ORM annotations.

PR #321, opened 2015-08-03
Feature/print

swinkelmann:
printservice: 
print coordinates, dynamic image and dynamic text

printclient: 
checkbox legend_default added
use current map scale-1 as default scale
admin type changed (+ button)


hwbllmnn:
Merged into develop, would be great if you'd also add a rebased version for the release-3.0.5 branch.

PR #330, opened 2015-08-11
printservice: bugfix coordinate print

swinkelmann:
bugfix print coordinates


hwbllmnn:
Merged manually because target branch was not release/3.0.5

PR #406, opened 2016-03-11
add optimized Export/Import/Copy

psch:


eSlider:
Nice work! Application copy runs 3 time faster then before!

Here two test with some large configured  application on my machine:

# Before merge

1. copydirectly 1.7 min
2. copydirectly 1.7 min	

# After merge

1. copydirectly 30.04s
2. copydirectly 28.22 s

--- 
Anything else works fine like before.

PR #416, opened 2016-04-27
fix scale, scaleHint, MinMax; clean/formate code

psch:


eSlider:
@AxxL, code is ok! But I can't check functionality change, can u?

AxxL:
OK

PR #424, opened 2016-05-27
No need to check datasource yet

debugteam:
No need to check datasource yet


eSlider:
@psch, do we need status check outside of mapbender3 himself, but any other mapbender3 based project?

PR #430, opened 2016-06-10
print: fixed template import with fpdi

swinkelmann:
removed FPDF-Plugin PDF_ImageAlpha.php, we dont need it anymore.
added PDF_Extensions.php with additional PDF functions.


eSlider:
@swinkelmann, PDF_ImageAlpha was removed by the request, but it's the any relations in mapbender based projects?

PR #431, opened 2016-06-10
An other workaround to disable autocomplete password (add wms source)

ThorstenHi:
Enhanced workaround for chrome based browsers.


LazerTiberius:
This will break the accessibility. The new input fields can get focus e.g. using the tab key.  But setting the input to  autocomplete="new-password" should work in in Chrome as well in Firefox.

PR #433, opened 2016-06-15
Fix print default font size setting

LazerTiberius:
Fix for application crashes if no fontsize is set. Now it will use default value 20 instead. @swinkelmann Please check before we can merge


AxxL:
I think you have to do the fix for the hotfix/print branch from @swinkelmann

PR #436, opened 2016-06-23
Update printclient.scss changed width to auto to support long titles

astroidex:
* labelinput is too small with 80px
* like for searchrouter width was set to auto


eSlider:
Thx, @astroidex!

PR #438, opened 2016-06-24
Feature/print bundle isolation

eSlider:
An example on how to extract print bundle from the mapbender core to own independent mapbender module. The code base of print bundle is moved here: https://github.com/mapbender/mapbender-printclient


AxxL:
@eSlider Can we delete that? Is that a test?

eSlider:
@AxxL, this is branch is a code extract of print service and element to an own repository. I move the pull-request to `release/3.0.7`

werrolf:
I really don't think this is the right time for this. There's a lot of work to do with print that is best done in tandem with the rest of the code.
* how print interacts with http (or owsproxy, right now)
* how print locates its resources (currently in mapbender-starter)
* how print misses to reuse ImageExport code
* how blissfully unaware print / image export are of Wms (e.g. no tiled requests, leading to unnecessarily high memory consumption, sometimes even WMS-service-side refusal)

The plan I see for print would both extract responsibilities and add new support components. Some of those new classes will naturally belong into mapbender core, others in WmsBundle. Splitting repositories just makes this work harder.

There's also the observation that almost all ongoing projects utilize print. It feels wrong to declare it optional looking at the real world applications of Mapbender.

The only thing that gets easier if we break out the print bundle as is would be forking and replacing it entirely.
I'd much, much prefer to resolve any apparent issues centrally.

AxxL:
Thanks. I have to close this.

PR #451, opened 2016-08-22
icon visibility of application in configuration #7082

AxxL:
please check.

-- 
send from my FOSS4G CodeSprint


eSlider:
Closes #451

PR #455, opened 2016-08-23
fixed featureinfo css in certain conditions

hwbllmnn:
fixes internal issue 5983


hwbllmnn:
Needs more testing, breaks some setups.

PR #456, opened 2016-08-23
Hotfix/redlining text

### LINKED ISSUE #401: Redlining: edit text does not chnage text color in the map and ancor is not visible
- redlining allows to change existing objects. for point, line, polygon the objects are changed from red to blue, when you click on button pen to edit the object

Bug
- button pen, does not change the color of the text to blue (when you have many textes you do not see which one is choosed)
- the ancor of the text is at the lower-left of the text. But you do not see it.

Solution
- change color of text that you want to edi to blue
- show the ancor point of the text (maybe as a point)

You can test it in demo;

http://version.mapbender3.org/3053/application/mapbender_user

eSlider:
Fix for #401 by @psch. 

@astroidex, @CharlyUnicorn: please make sure this pull-request resolves #401 ticket.

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
@@ -17,7 +17,7 @@
         layer: null,
         activeControl: null,
         selectedFeature: null,
-        geomCounter: 1,
+        geomCounter: 0,
         rowTemplate: null,
         _create: function(){
             if(!Mapbender.checkTarget("mbRedlining", this.options.target)) {
@@ -176,10 +176,10 @@
                                 featureAdded: function (e) {
                                     if ($('input[name=label-text]', self.element).val().trim() === '') {
                                         Mapbender.info(Mapbender.trans('mb.core.redlining.geometrytype.text.error.notext'));
-                                        self.removeFeature(e);
+                                        self._removeFeature(e);
                                     } else {
                                         e.style = self._generateTextStyle($('input[name=label-text]', this.element).val());
-                                        self._addToGeomList(e, Mapbender.trans('mb.core.redlining.geometrytype.text'));
+                                        self._addToGeomList(e, Mapbender.trans('mb.core.redlining.geometrytype.text.label'));
                                         self.layer.redraw();
                                         $('input[name=label-text]', this.element).val('');
                                     }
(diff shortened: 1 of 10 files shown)


werrolf:
I'd prefer to merge this into 3.0.5 because it seems like the use case is legacy applications that cannot use the new mapbender-digitizer.

astroidex:
we do not need it in MB 3.0.5. We only use 3.0.6 now and in the future

astroidex:
If redlining is obsolete, then we do not need this hotfix anymore.
 
Quotation from above
"CoreBundle\Element\Redlining is obsolete / superseded by mapbender-digitizer, but LGTM."

AxxL:
Fix for 3.0.6.

werrolf:
Repeated binding / unbinding the keyup event on the input[name='label-text'] seems inefficient, even though the event handler can not be bound several times for repeated _modifyFeature calls (the call to _deactivateControl unbinds the event first).

Event on here: https://github.com/mapbender/mapbender/blob/4642762d7fdbfc73a6e6702f69891117879d155c/src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js#L261
Event off here: https://github.com/mapbender/mapbender/blob/4642762d7fdbfc73a6e6702f69891117879d155c/src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js#L205

The event handling method _writeText already seems to verify (indirectly, by checking for a label in the feature style) that the currently selected feature is of type text. So it may be safe (and clearer) to consolidate event bind / unbind in the activate and deactivate methods, alongside the click event on .redlining-tool.

This should also be a change event, not keyup. Keyup is not always generated on copy&paste interactions and is never generated on form reset.


I also noticed (though not related to changes in this pull) that _newControl will happily leak Openlayers Control objects, which could and should be simply reused per type.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
```diff
@@ -184,33 +191,49 @@
             this.activeControl.activate();
 
         },
-        removeFeature: function(feature){
+        _removeFeature: function(feature){
```
eSlider: Public method change to private in current bug-fix release isn't so cute. :)

src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
```diff
@@ -184,33 +191,49 @@
...
             this.layer.destroyFeatures([feature]);
         },
+        _removeAllFeatures: function(){
+            $('.geometry-table tr', this.element).remove();
+            this.map.removeLayer(this.layer);
+        },
         _deactivateControl: function(){
+            if(this.selectedFeature) {
+                this.activeControl.unselectFeature(this.selectedFeature);
+                if (this.selectedFeature.style && this.selectedFeature.style.label) {
```
eSlider: What if this isn't this, but that or me and my self?

src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
```diff
@@ -242,24 +271,41 @@
...
         },
-        _setFeatureStyle: function(){
+        _generateTextStyle: function(label){
             var style = OpenLayers.Util.applyDefaults(null, OpenLayers.Feature.Vector.style['default']);
-            var label = $('input[name=label-text]', this.element).val();
-            style.label = label;
+            if (label) {
+                style.label = label;
+            }
             style.labelAlign = 'lm';
```
eSlider: Put magic numbers into `options`!

eSlider: As default and merge it:  
```js

// Default widget style option
// Surely it's need to be defined on top of the widget.
widget.options.style = {labelXOffset:10, pointRadius: 6};

// And now we merge defaults to style object. Pure magic, not? :)
_.extend(style,widget.options.style);
```

PR #474, opened 2017-01-21
fix duplicate loads of WMS

hwbllmnn:
Fixes duplicate loads of WMS when a layer is going out of scale.

Fixed internal issue 7689.


eSlider:
Thank you, @hwbllmnn!

PR #480, opened 2017-02-01
Hotfix/basesourceswitcher initialization

wg-rolf:
Fixes weird / wasteful base source requests on currently invisible base sources when embedding BaseSourceSwitcher.
With an active BaseSourceSwitcher, only one layer may be active, but previously, all base sources would start loading anyway, which is wasteful on more extreme use cases of the BSS (some people use 60+ base sources). Also, depending on the layer source nesting, some requests may go out with LAYERS= empty, which can cause secondary issues with proxies.

Fix is built on a (small) extension to Application + Element classes, which allow server-side Application config modifications by active Element instances.

This might be a better approach than previous purely client-side fixes to e.g. SuggestMap Elements, where the common root of the problem is that the application is initialized into a certain "default" state on the client, and then that configuration is immediately altered. Working around event order dependencies with that model is challenging. Amending the initial configuration server-side allows much easier, more robust fixes.

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Application.php
+++ b/src/Mapbender/CoreBundle/Component/Application.php
@@ -352,6 +352,18 @@ class Application
                 }
             }
         }
+        // Let (active, visible) Elements update the Application config
+        // This is useful for BaseSourceSwitcher, SuggestMap, potentially many more, that influence the initially
+        // visible state of the frontend.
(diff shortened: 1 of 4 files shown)


wg-rolf:
Fix is incomplete, depending on BSS configuration, unwanted requests can still go out to deactivated sources. Work on the branch will continue to fix all remaining cases.

PR #481, opened 2017-02-02
print and imageexport with content-type image/png8 and image/png mode…

swinkelmann:
…=24bit


eSlider:
Well done, @swinkelmann! :+1:

Issue #477, opened 2017-02-01
PrintClient - print legend on first page


astroidex:
You can define a legend region on the first page along with your map. This region will be used to print the legend. when the region is full a second page will be used for the legend

eSlider:
@astroidex, Is merged!

CharlyUnicorn:
if you enable multiple layers/wms and start a print with the legend on the first page the legend is printet on the first page. Due to the lack of space, the remaining images are printed on the second page.

Example with multiple mapbender_user and germany wms:  
![auswahl_119](https://user-images.githubusercontent.com/13517141/29863647-a0abadce-8d70-11e7-9271-3aab16c45fb5.png)

Problem: If you load only one wms multiple times, then the  remaining images are not printed on the second page. They dissapear at the end of the first page. 
Example with multiple mapbender_user wms (germany wms disabled): 
![auswahl_120](https://user-images.githubusercontent.com/13517141/29863643-9e31ee46-8d70-11e7-9931-8568418e6c14.png)

It would be good if no legend images were lost on the first page, but i'm not sure if this is a bug or a feature. The question is when you have to load a wms service mutliple times, so that it comes to a problem here...

AxxL:
It's not intended to work this way. I guess it's ok for @astroidex projects to work like this and every change in behaviour would be a CR regarding this functionality. We cannot "fix" this behaviour right now if the requirements haven't placed that issue. So, unfortunetaly right now: Works as desiged, missing ressources, low priority.

Issue #476, opened 2017-02-01
PrintClient - print a logo on the legend page


astroidex:
define a logo in the print odg file  and provide a logo image. The logo will then be printed on the logo page at the defined area

astroidex:
Have a look at the pull request but please add it to release 3.0.6

astroidex:
https://github.com/mapbender/mapbender/pull/462

PR #486, opened 2017-03-02
legendpage_image and logo on legend page #476 #477

astroidex:
* see Tickets #476 #477
* see also changes in mapbender-starter


eSlider:
Thank you, @astroidex!

PR #496, opened 2017-03-30
fixed typos, merged with en version, changed some translation

astroidex:
* merged the structure with the en version. same ordering
* fixes some typos f.e. sidepane Buttons instead of Tabs (as Reiter as Reiter is the wŕong translation)


src/Mapbender/ManagerBundle/Resources/translations/messages.de.yml
```diff
@@ -212,14 +208,14 @@ mb:
     template:
       region:
         tabs:
-          label: Reiter
+          label: Buttons
```
AxxL: Tabs = Tabs

PR #506, opened 2017-04-06
Reenable default text selection in nested elements in map

wg-rolf:
Layer tiles remain non-selectable.
Text inside Openlayers popups becomes selectable again (previously not possible).
Solved by pushing user-select attribute change down from .mb-element-map to a nested .olLayerDiv selector.

This fixes internal ticket #8068.


eSlider:
Thx, @wg-rolf

Issue #507, opened 2017-04-07
map Element - parse icon/size/offset as url-parameter on application start


astroidex:
Mapbender can be started with different url parameter: 
http://doc.mapbender3.org/de/bundles/Mapbender/CoreBundle/elements/map.html#kontrolle-uber-url-parameter

You can parse pois with coordinates and labels. But you can't define an icon for a poi

Enhancement: define new parameters
* poi[0][icon]=http://localhost/grillplatz.png -> Url zu einem Icon
* poi[0][icon_width]=40& ->  width & height
* poi[0][icon_height]=40
* poi[0][icon_xoffset]=20 -> Offset  Icon
* poi[0][icon_yoffset]=20

astroidex:
* branch for mapbender 3.0.5 https://github.com/mapbender/mapbender/tree/feature/map-getparameter-poi-icon
* have a look at pullrequest for release/3.0.5 https://github.com/mapbender/mapbender/pull/508

astroidex:
Request could be
http://localhost/mapbender/application/mapbender_user?poi[0][point]=363374,5621936&poi[0][label]=Label&poi[0][icon]=http://localhost/mapbender/bundles/mapbendercore/image/grillplatz.png&poi[0][icon_width]=40&poi[0][icon_height]=40&poi[0][icon_xoffset]=20&poi[0][icon_yoffset]=20

Could you please check whether the url that is pared could cause trouble?

AxxL:
I'm closing this issue now since no progress in PR. Feel free to open again if the PR is ongoing and we can check the feature.

PR #508, opened 2017-04-07
map enhancement, add url parameter icon #507

astroidex:
* this pull request is for branch release/3.0.5
* see ticket https://github.com/mapbender/mapbender/issues/507

### DIFF
--- a/src/Mapbender/CoreBundle/Element/Map.php
+++ b/src/Mapbender/CoreBundle/Element/Map.php
@@ -162,7 +162,12 @@ class Map extends Element
                     'x' => floatval($point[0]),
                     'y' => floatval($point[1]),
                     'label' => isset($poi['label']) ? htmlentities($poi['label']) : null,
-                    'scale' => isset($poi['scale']) ? intval($poi['scale']) : null
+                    'scale' => isset($poi['scale']) ? intval($poi['scale']) : null,
+                    'icon' => isset($poi['icon']) ? $poi['icon'] : null,
+                    'icon_width' => isset($poi['icon_width']) ? intval($poi['icon_width']) : null,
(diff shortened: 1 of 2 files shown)


src/Mapbender/CoreBundle/Element/Map.php
```diff
@@ -162,7 +162,12 @@ public function getConfiguration()
                     'x' => floatval($point[0]),
                     'y' => floatval($point[1]),
                     'label' => isset($poi['label']) ? htmlentities($poi['label']) : null,
-                    'scale' => isset($poi['scale']) ? intval($poi['scale']) : null
+                    'scale' => isset($poi['scale']) ? intval($poi['scale']) : null,
+                    'icon' => isset($poi['icon']) ? $poi['icon'] : null,
```
eSlider: @LazerTiberius, please take a look on this.

LazerTiberius:
I tried different possible exploits but without success. OL2 seems to do a good job here. The only problem I see is the possibility to load any file with the parameter. But this is more a problem with the general design. Maybe we should consider to only allow predefined icons and only use a name/id as a parameter?

eSlider:
@LazerTiberius, if this good, please mark as approved. thx!

LazerTiberius:
> But this is more a problem with the general design. Maybe we should consider to only allow predefined icons and only use a name/id as a parameter?

There is still an open question from me?

AxxL:
Since this is a feature the pull-request has to be done against the 3.0.7 branch.

@eSlider @LazerTiberius I don't think using anything than predefined icons is a security thread and should not be put into mapbender without validating the concerns. If remote icons should be displayed, the core must be hardened to avoid loading remote JS-code or anything else except for an icon.

 @astroidex @eSlider As this is a feature, it must be a pull request against 3.0.7.

PR #532, opened 2017-05-17
improved position of removebutton

WGSZissner:
For better allocation, changed removebutton position to right


AxxL:
Which button? Which test-szenario? Does a ticket exist?

Issue #525, opened 2017-05-10
Print: Specific WMS cannot be printed


AxxL:
Specific WMS with raster data cannot be printed. Print service logs cryptic Print-URL. Maybe problem in the Capabilities?

See internal ticket 8185 for URL.

AxxL:
Test running for:

"mapbender/mapbender": "dev-hotfix/print-content-type as 3.0.6.1",

- JPG, GIF
- Export Map.

swinkelmann:
@AxxL please test it again. Should also work with jpeg and gif.

AxxL:
OK thanks.

PR #540, opened 2017-05-22
PrintService/ImageExport: Accept all kinds of image/png from wms

swinkelmann:
Fixes issue: #525


swinkelmann:
#525 PrintService/ImageExport: Accept all kinds of image/jpeg and image/gif from wms

PR #542, opened 2017-05-26
Fixing issues with WmsLayerSource::getAuthority

werrolf:
1) Second path in the getter's if statement (``$inherit=false`` or no parent or no authorities set on parent) was missing the return statement; implied return has been null, even in layer sources where an authority (or multiple) existed
2) We support multiple authorities per layer source. The getter returns an array of Authority objects, preinitialized to an empty array in the constructor. All modification uses array operations. All setting sites use array format.
Hence the getter/setter pair should be named in the plural, ``getAuthorities`` & ``setAuthorities``, to avoid confusion.
3) The getter's code path for ``$inherit=true`` has made an effort to distinguish empty arrays from null. The only way for the parent authority to be null however is issue 1 above. To be safe against malformed data in a legacy installation db, we fall back to boolean coersion, which treates empty arrays and nulls the same with less code.


WmsBundle/Entity/WmsLayerSource.php
```diff
@@ -770,28 +770,50 @@ public function addAuthority(Authority $authority)
     }
 
     /**
-     * Set authority
+     * Set authoritIES
```
atd-schubert: Cap lock???

werrolf:
@AxxL the branch target was supposed to be 3.0.5, but even local logs show it was based on the wrong parent.
Technically it's an API change (though fully BC, it adds methods), so you could argue that it needs a minor version bump, not just patch-level.

I'll close this for now. We can take the changes (minus spelling details) into a new branch based on 3.0.6.

Issue #543, opened 2017-05-30
Search Router - OpenLayers Error in application after update  to version 3.0.6.1


CharlyUnicorn:
I updated multiple Mapbender3 from version 3.0.5.3 to version 3.0.6.1. 
I used the search router element in my applications without any problems.
After the update I opened the application and couldn't use any functions related to the map-element/ OpenLayers.

The following error pops up: 

```
unreachable code after return statement
TypeError: OpenLayers is undefined

further Information Link: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Unexpected_type?utm_source=mozilla&utm_medium=firefox-console-errors&utm_campaign=default
```

- If I remove the searchrouter the application works. 
- If I delete the searchrouter and add a new searchrouter element the error still shows up.  
- If I update the existing element the error still shows up.
- If I create a new application or copy an existing application (with the error) the application works.

![auswahl_042](https://cloud.githubusercontent.com/assets/13517141/26587337/52b4a974-4552-11e7-8287-4e3bf96e78db.png)

CharlyUnicorn:
perhaps it would be helpful, if you could load the assets from the map and then the assets from the searchrouter

CharlyUnicorn:
new error workflow: 
if you have an application with the searchrouter and the digitizer then the application works after the update - but if you change something at the digitizer-yaml, then the same error pops up.

CharlyUnicorn:
You can test this with the following application: 
* https://demo.mapbender3.org/application/mapbender_user_reference

eSlider:
@RobinSchwammborn, can u check this?

RobinSchwammborn:
I checked it, critical bug has been fixed.

Urbai:
how to set up seach router in mapbender3 when we have for SGBD postgresql?

Urbai:
Good morning every body,  how to set up seach router in mapbender3 when we have for SGBD postgresql?

Urbai:
Hello, please help me.
 how to set up seach router in mapbender3 when we have for SGBD postgresql?

astroidex:
Hello,

please update to https://mapbender.org/builds/3.0.6.3/ not 3.0.6.3

Urbai - please better send a mail to the mailing list. You can register here: https://lists.osgeo.org/mailman/listinfo/mapbender_users

Urbai:
ok thank you, we are going to try it.

PR #544, opened 2017-05-30
patch for ticket #543: added new functionality for elements to load a…

WGSZissner:
…ssets late
- have a look at ticket https://github.com/mapbender/mapbender/issues/543

this hotfix should solve the problem

i added a new method for Elements to load assets later name getLateAssets in Component/Element.php


LazerTiberius:
I don´t think this is a solution for the real problem, more a workaround which will fix it just for the moment. We should take a deeper look into this bug. Normally OpenLayer should be loaded a long time before any element gets initialized.

Issue #530, opened 2017-05-16
element map - Open an Application with poi speech bubble is not shown


astroidex:
* when you open an Application with poi speech bubble is not shown for existing applications
* for applications that were created in 3.0.6.0 the icon is there 
* see internal ticket #7978

* works fine with new applications https://demo.mapbender3.org/application/mapbender_user?poi[point]=366164%2C5623183&poi[scale]=25000&poi[label]=Please+take+a+look+at+this+POI%3A

* DOES NOT WORK with application that was created in 3.0.5.3
https://demo.mapbender3.org/application/mapbender_digitize_demo?poi[point]=366164%2C5623183&poi[scale]=25000&poi[label]=Please+take+a+look+at+this+POI%3A 

* missing icon: https://demo.mapbender3.org/bundles/mapbendercore/mapquery/lib/openlayers/img/cloud-popup-relative.png

astroidex:
I found the reason: Map-Element: Image path changed and is not changed in the configuration in the Database
* old imgPath: bundles/mapbendercore/mapquery/lib/openlayers/img
* new imgPath: components/mapquery/lib/openlayers/img

How can you fix it:

1.  put the image to the old path or create a symbolic link
2. change the imgPath for every map-element to the new path - you can do this via administration-backoffice in the browser or via SQL

For 2 via SQL. you could use the following SQL
```
UPDATE mb_core_element
   SET configuration=replace(configuration,'"imgPath";s:49:"bundles/mapbendercore/mapquery/lib/openlayers/img"','"imgPath";s:38:"components/mapquery/lib/openlayers/img"') 
WHERE class= 'Mapbender\CoreBundle\Element\Map' ;
```
This is only a workaround. A solution for existing applications would be great

AxxL:
Vorschlag: Write an app/console command (mapbender:database:upgrade) was diesen Wert aktualisiert. Ich schlage vor, nicht in SQL, sondern mit Doctrine mitteln.

Das Kommando müssen wir dann für weitere Upgrades von DB-Inhalten erweitern können.

LazerTiberius:
I have done some minor changes to @WGSZissner  commit,  see pull request #545 . Needs still a review from a third person @wg-rolf  @eSlider

AxxL:
This command doesn't give me any hint.
@WGSZissner

AxxL:
@LazerTiberius help out please.

eSlider:
[...]

PR #545, opened 2017-05-31
Added Command to update old imagepath of map element #530

WGSZissner:
Command/DatabaseUpgradeCommand.php

will change Configuration.imgPath of each MapElement from "bundles/mapbendercore/mapquery/lib/openlayers/img" to "components/mapquery/lib/openlayers/img"


LazerTiberius:
Like we discussed on Friday I added some output for the UX

PR #551, opened 2017-06-29
remove 'move' check on click event

### LINKED ISSUE #548: SearchRouter in mobile application - AutoClose after click on result
If you create an application with the "Mapbender Mobile template" you can add a search via the elements SearchRouter or SimpleSearch. 
- Correct: If you search with the SimpleSearch (Solr) you can click on the result in the list and the window closes automatically. Then you can see the entry on the map. 
- Wrong: If you search with the SearchRouter you can click on the result in the list and the window doesn't closes automatically. Nothing shows you, that you triggered an event like the zoom to the entry. Then you can't see the entry on the map until you manually close the window. 

You can test this with this demo application: 
- https://demo.mapbender3.org/application/suche_test_mobil
- Search: "Emde"

WGSZissner:
fixes #548

edited event listener for searchroutrer result click only on mobile template.

now the search overlay closes if the user clicks an an search-result

### DIFF
--- a/src/Mapbender/MobileBundle/Resources/public/js/mapbender.mobile.js
+++ b/src/Mapbender/MobileBundle/Resources/public/js/mapbender.mobile.js
@@ -115,9 +115,8 @@ $(function(){
         });
     });
 
-    $('.search-results').on('click', function(){
-        if(moved){
-            moved = false;
+    $('.search-results').on('click', function(e){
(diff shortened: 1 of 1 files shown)


astroidex:
cool. thank you.

eSlider:
Closes: #548

PR #576, opened 2017-07-19
Hotfix/translations

WGdogan:


astroidex:
* @eslider please delete this pull request. We will merge the whole hotfix/translations when everything is finished. So better later. This was only for the turkish translation from @WGdogan

Issue #584, opened 2017-07-26
WMS with Scale fails to load


AxxL:
WMS-URL: http://inspire.brandenburg.de/services/strassennetz.php?SERVICE=WMS&REQUEST=GetCapabilities

Steps to reproduce:
* Load WMS as datasource (works)
* Load WMS into application.

Expected result: Layerset-Instance dialog shows up.

Actual result: Fails to load with following Error:

```
Call to a member function getMin() on null
500 Internal Server Error - FatalThrowableError
Stack Trace

    in mapbender/src/Mapbender/WmsBundle/Entity/WmsLayerSource.php at line 646  -
                }
                if ($recursive && $value === null && $this->getParent()) {
                    $value = $this->getScaleRecursive()->getMin();
                }
                $value === null ? null : floatval($value);

```

AxxL:
/ref #565 (Reference)

http://wms2016.wheregroup.com/cgi-bin/mapserv?map=/data/umn/test/test.map&Version=1.3.0&Service=WMS&REQUEST=getCapabilities

AxxL:
Pull Request #585

PR #585, opened 2017-07-27
Hotfix/load wms get min on null

LazerTiberius:
Hotfix for WMS with Scale fails to load #584

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+* **v.3.0.6.3**
+    - Fix WMS with Scale fails to load #584 - see commit message #2783540 for more information
+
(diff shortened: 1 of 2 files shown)


wg-rolf:
getScaleRecursive does not return anything if (local) scale is an object with min and max set and the instance has no parent to recurse to.

I.e. missing ``else return $scale`` on this if:
https://github.com/mapbender/mapbender/blob/d25eaab8abe5c8ccaca0c9c7be1cec8bdf75422c/src/Mapbender/WmsBundle/Entity/WmsLayerSource.php#L695

eSlider:
@LazerTiberius, are you core with @wg-rolf? 
@wg-rolf, can I merge this?

PR #606, opened 2017-08-22
Refactor legend image element

eSlider:
Legend image refactored and improved:
* Split code in logical parts (methods)
* Removed unnecessary repeated code
* Added method documentation
* Images loaded first if requiered, not all at once
* Every time render completely new HTML, also if map layers visibility changed
* Add simple progress loader (spinner) this will displayed in HTML container before images was loaded  
* Fixed open/close/ready event handling (callbacks)
* Fixed posible inline HTML XSS 
* Using jQuery instead of String-HTML concatination
* Render HTML only first if Map is ready

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
@@ -136,6 +136,7 @@
                     closeButton:            false,
                     closeOnPopupCloseClick: true,
                     closeOnESC:             false,
+                    destroyOnClose:         true,
                     content:                (html),
                     width:                  350,
                     height:                 500,
@@ -431,14 +432,19 @@
          */
         close: function() {
             var widget = this;
-            if(widget.isPopUpDialog) {
-                if(widget.popup) {
-                    if(widget.popupWindow.$element) widget.popupWindow.destroy();
-                    widget.popupWindow = null;
+
+            if (widget.isPopUpDialog) {
+
+                if (widget.popup) {
+
+                    if (widget.popupWindow.$element) {
+                        widget.popupWindow.destroy();
+                        widget.popupWindow = null;
+                    }
                 }
 
             }
-            if(widget.callback) {
+            if (widget.callback) {
                 widget.callback.call();
                 widget.callback = null;
             }
@@ -474,7 +480,7 @@
             widget.readyState = true;
             
             // Remove handlers
-            readyCallbacks.splice(0, readyCallbacks.length);
+            widget.readyCallbacks.splice(0, widget.readyCallbacks.length);
 
         }
 
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
@@ -19,7 +19,6 @@
 
         readyCallbacks: [],
         callback:       null,
-        model:          null,
 
         /**
          * Widget constructor
@@ -43,10 +42,9 @@
         _setup: function() {
             var widget = this;
             var options = widget.options;
-            var mapWidget = $("#" + widget.options.target).data("mapbenderMbMap");
+
 
             options.noLegend = Mapbender.trans("mb.core.legend.nolegend");
-            widget.model = mapWidget.getModel();
 
             // Deprecated check if options exists
             if(options.hasOwnProperty("showGrouppedTitle")) {
@@ -165,7 +163,7 @@
         _getSources: function() {
             var widget = this;
             var allLayers = [];
-            var sources = widget.model.getSources();
+            var sources = Mapbender.Model.getSources();
             for (var i = (sources.length - 1); i > -1; i--) {
                 allLayers.push(widget._getSource(sources[i], sources[i].configuration.children[0], 1));
             }
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
@@ -1,515 +1,478 @@
-(function($){
+(function($) {
 
     $.widget("mapbender.mbLegend", {
         options: {
-            autoOpen: true,
-            target: null,
-            noLegend: "No legend available",
-            elementType: "dialog",
-            displayType: "list",
-            checkGraphic: false,
-            hideEmptyLayers: true,
+            autoOpen:                 true,
+            target:                   null,
+            noLegend:                 "No legend available",
+            elementType:              "dialog",
+            displayType:              "list",
+            checkGraphic:             false,
+            hideEmptyLayers:          true,
             generateLegendGraphicUrl: false,
-            showSourceTitle: true,
-            showLayerTitle: true,
-            showGrouppedTitle: true,
-            maxImgWidth: 0,
-            maxImgHeight: 0
+            showSourceTitle:          true,
+            showLayerTitle:           true,
+            showGroupedTitle:         true,
+            maxImgWidth:              0,
+            maxImgHeight:             0
         },
-        model: null,
-        layerTitle: "",
-        sourceTitle: "",
-        grouppedTitle: "",
-        hiddeEmpty: "",
-        _create: function(){
-            if(!Mapbender.checkTarget("mbLegend", this.options.target)) {
+
+        readyCallbacks: [],
+        callback:       null,
+        model:          null,
+
+        /**
+         * Widget constructor
+         *
+         * @private
+         */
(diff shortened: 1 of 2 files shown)


src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
```diff
@@ -1,515 +1,478 @@
...
+         * On ready handler
          */
-        _ready: function(){
-            for(callback in this.readyCallbacks) {
-                callback();
-                delete(this.readyCallbacks[callback]);
+        _ready: function() {
+            var widget = this;
+            var readyCallbacks = widget.readyCallbacks;
+            for (var k in readyCallbacks) {
```
atd-schubert: We should not use `for-in` loops on arrays!

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
```diff
@@ -1,515 +1,478 @@
...
+                var readyCallback = readyCallbacks[k];
+                if(typeof(readyCallback ) === 'function') {
+                    readyCallback();
+                }
             }
-            this.readyState = true;
+            // Remove handlers
+            readyCallbacks.splice(0, readyCallbacks.length);
+            widget.readyState = true;
```
atd-schubert: The `ready` method proofs first that `widget.readyState` is set to `true` and if it is so it runs the callback function directly, maybe we should here also set the value of `readyState` to `true` first and then  call all the registered callbacks.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
```diff
@@ -1,515 +1,478 @@
...
-        _ready: function(){
-            for(callback in this.readyCallbacks) {
-                callback();
-                delete(this.readyCallbacks[callback]);
+        _ready: function() {
+            var widget = this;
+            var readyCallbacks = widget.readyCallbacks;
+            for (var k in readyCallbacks) {
+                var readyCallback = readyCallbacks[k];
+                if(typeof(readyCallback ) === 'function') {
```
atd-schubert: Instead of proofing that it is a function we can add a `try catch` before we run the loop, or the forEach function. With this way, we can handle errors from the callbacks also...

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
atd-schubert: I would prefer
```js
readyCallbacks.forEach(function(cb) {
  cb();
})
```

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
```diff
@@ -1,515 +1,478 @@
...
+                    readyCallback();
+                }
             }
-            this.readyState = true;
+            // Remove handlers
+            readyCallbacks.splice(0, readyCallbacks.length);
```
atd-schubert: If we use the `forEach` method we have to reset the readyCallbacks. Something like that:

```js
widget.readyCallbacks = []
```

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
```diff
@@ -1,515 +1,478 @@
...
+        /**
+         * Widget constructor
+         *
+         * @private
+         */
+        _create: function() {
+            var widget = this;
+            var options = widget.options;
+            if(!Mapbender.checkTarget("mbLegend", options.target)) {
                 return;
```
atd-schubert: Don't we have to throw an Error here?

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
```diff
@@ -1,515 +1,478 @@
...
+
+        /**
+         * Setup widget
+         *
+         * @private
+         */
+        _setup: function() {
+            var widget = this;
+            var options = widget.options;
+            var mapWidget = $("#" + widget.options.target).data("mapbenderMbMap");
```
atd-schubert: I'm not happy to search the target by an ID string instead of using the HTMLElement directly, but I think this is a deeper change to the architecture.

But we should keep in mind, that we should refactor this.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
atd-schubert: Maybe in addition to the `$.notify` function

src/Mapbender/CoreBundle/Resources/views/Element/legend.html.twig
```diff
@@ -1,4 +1,4 @@
 <div id="{{ id }}" class="mb-element mb-element-legend" title="{{ title |trans }}">
-    <div id="imgtest"></div>
-    <div id="legends"></div>
+  <div id="imgtest"></div>
```
atd-schubert: What is the purpose of `imgtest`? Why do we pollute the DOM IDs with that?

philvanbwg:
I tested this Code with Mapbender 3.0.6.3: It works in both regions, sidepane and topbar. It works when a wms is activated in the Layertree as well as in the basesourceswitcher. It also works when a WMC is loaded or being edited! This should be the most common use cases. As far as I can tell this solution works just fine.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
eSlider: @atd-schubert, as we discuss about how-to reset callbacks and steel have same array reference, I let this code part unchached

CharlyUnicorn:
Hi, 
I tested the changes and had some problems: 

1. overwrites the changes from https://github.com/mapbender/mapbender/pull/609
2. legend dialog opens if you enable/disable layer in the layertree

Workflow for tests 1: 
- add changes from pull-request 609
- open the legend element in the sidepane 
- you can't see the slider from the change "overflow: auto"

Workflow for tests 2: 
- open the legend via a button in a popup-dialog
- close the dialog
- enable/disable a layer/ wms in the layertree
- bug: dialog for the legend opens with every click in the checkbox (en/disable)

sveta-wl:
Fixed the issue "legend dialog opens if you enable/disable layer in the layertree".

philvanbwg:
Svitlana's commit fixes the bug mentioned by Charly. I tested it and everything works. This is good to go now.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
AxxL: @LazerTiberius Check, if the map is available? Scoping?

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
AxxL: @LazerTiberius Do it.

Please configure target for the element.

Bitte konfigurieren Sie das Target für das Element.

LazerTiberius: > Don't we have to throw an Error here?
  Actually, checkTarget will display an error in the right corner and also one in the js console if no target is defined.

AxxL:
@atd-schubert REVIEW YOUR REVIEW!

src/Mapbender/CoreBundle/Resources/views/Element/legend.html.twig
```diff
@@ -1,4 +1,4 @@
...
+  <div id="legends" class="legends"></div>
```
atd-schubert: Why do we have a class and an id here? I know that the comment above about `imgtest` has nothing to do with this change, it was just a change of the indent, but in this case I think we could really refactor it and remove the DOM id!

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
atd-schubert: When we refactor this with `var`s for the variables, we should also refactor the `for in` loop!

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
atd-schubert: 👍

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
atd-schubert: @LazerTiberius this is the important change request...

PR #609, opened 2017-08-22
CSS-Fix for large Legend images

philvanbwg:
Just added a little css to show a scrollbar for oversized legend images that were cut off before (see screenshot). Changes are in the legend.scss file.
And removed the mapbender.element.legend.css file since it doesn't seemed to be used anymore.

2017-07-20-legende_layertree


src/Mapbender/CoreBundle/Resources/public/sass/element/legend.scss
```diff
@@ -1,4 +1,10 @@
 
 // notshow -> hide always legend item
 // notvisible -> hide legend item by user
-.notshow, .notvisible{display:none;}
\ No newline at end of file
+.notshow, .notvisible{display:none;}
+
+.mb-element-legend ul li.image {
+
+  overflow: scroll;
```
RobinSchwammborn: I would suggest "overflow: auto". It's just a small design tweak, but the scrollbars disappear if an element doesn't need them. That reduces the legend in many cases.
Alternatively, we could change the whole approach and change the image to a smaller scale than the sidepane-width (e.g. max-width: 200 px). Due to possible inconsistencies that can occur because of the size of different legend graphics and the adjustability of the sidepane, I don't quite know which solution fits more here.

philvanbwg: Of course, you are right on that. It needs to be "overflow: auto", my mistake.

Your second point is good also. This could be easily achieved by adding the bootstrap-css-class "img-responsive" to the -elements. I tested this and it works. But the readability of some scaled down images may suffer from that, so I rather prefer using "overflow:auto"

CharlyUnicorn:
Hi, 
I tested the changes and it works fine with my wms. The scrollbar for oversized legend images appears in the sidepane and popup element. 
Tested with the following wms: 
http://wms.wheregroup.com/cgi-bin/mapserv62?map=/data/umn/germany/germany_group.map&REQUEST=GetCapabilities&SERVICE=WMS&VERSION=1.3.0


auswahl_117

src/Mapbender/CoreBundle/Resources/public/sass/element/legend.scss
```diff
@@ -1,4 +1,10 @@
 
 // notshow -> hide always legend item
 // notvisible -> hide legend item by user
-.notshow, .notvisible{display:none;}
\ No newline at end of file
+.notshow, .notvisible{display:none;}
```
atd-schubert: You should indent `display: none` in a new line...

src/Mapbender/CoreBundle/Resources/public/sass/element/legend.scss
```diff
@@ -1,4 +1,10 @@
...
+
+.mb-element-legend ul li.image {
+
```
atd-schubert: Remove unnecessary blank line

AxxL:
Ich wurde von @werrolf dazu gezwungen!

AxxL:
Aber ich lasse das auch mit mir machen.

PR #610, opened 2017-08-22
Enable dimension handler

eSlider:
There is an Dimension Handler realised in some local project, avaible as source code, but never enabled. I suggest enable it in `release/3.0.7`


src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -956,10 +956,28 @@
                         lastItem.after(chkbox);
                         var inpchkbox = chkbox.find('.checkbox');
                         inpchkbox.data('dimension', item);
+                        initCheckbox.call(inpchkbox);
```
atd-schubert: Where is this global function from? Can't we require or import this in a "normal" way?

We use this global function also in:
* application-edit.js
* mapbender.element.wmceditor.js

LazerTiberius: This function is defined in FOM/CoreBundle/Resources/public/js/widgets/checkbox.js in the global scope. In my opinion, we should move this to some kind of utility object or something similar. Defining something in the global scope without a namespace is always bad.

LazerTiberius: I created an issue regarding this global ui helper functions https://github.com/mapbender/mapbender/issues/615

werrolf:
NTS1: if this doesn't meet acceptance (documentation, usability), we can trivially un-expose the element with a single-line change in src/Mapbender/WmsBundle/MapbenderWmsBundle.php
NTS2: if we do accept we need a changelog entry

PR #616, opened 2017-09-08
Fix tunnel getlegendgraphic

werrolf:
GetLegendGraphic for protected WMS now uses the same path in instanceTunnelAction as GetMap / GetFeatureInfo. "Special" bypass for legend urls has been removed.

### DIFF
--- a/src/Mapbender/CoreBundle/Controller/ApplicationController.php
+++ b/src/Mapbender/CoreBundle/Controller/ApplicationController.php
@@ -15,6 +15,7 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\Security\Acl\Domain\ObjectIdentity;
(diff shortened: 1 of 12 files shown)


werrolf:
@AxxL we should also merge this up into 3.0.6 + 3.0.7 ASAP after review.

werrolf:
Still problematic on this branch:
* adding a protected source to an application throws an error, source instance (=parent of the layer) id seems not (yet) initialized at ~L315, returns null:
```'instanceId' => $this->entity->getSourceInstance()->getId(),```
This leads to error in URL generation.

* it still transmits an (ineffective) legendurl param, which should be removed (L316); this doesn't really do anything but already confused us significantly in testing :)

**Comment by werrolf** [comment]
I think I have a working solution now.
I broke out a new method WmsInstanceEntityHandler::getRootLayerConfig in the NFC commit 56ce006. This returns the whole frontend-specific layer configuration array.

In the following commit I simply moved the call (and assignment into the config) from generateConfiguration (determines persisted DB values) into getConfiguration (returns client-facing configuration).

I tested with
* a password-protected WMS, I reloaded an existing one, I created a new one, works
* an unprotected WMS, same
* a YAML application with YAML-defined WMS sources

**Comment by werrolf** [comment]
For extreme hazard testing, please use the POC branch based on this one:
https://github.com/mapbender/mapbender/tree/fix/tunnel-getlegendgraphic-POC-pull616

This strips the entire layerset configuration ("children") from the DB-side configuration in WmsInstance in both the getter and the setter. WMS still functions as before. This demonstrates that persisting this data is not needed. I am not going to do a pull request on the POC branch. It will be deleted and forgotten after we have made a decision on this one.

File: src/Mapbender/CoreBundle/Controller/ApplicationController.php [hunk 1]
```diff
@@ -366,26 +375,28 @@ public function instanceTunnelAction($slug, $instanceId)
...
-                        $url = $instance->getSource()->getGetMap()->getHttpGet();
-                    } elseif (strtolower($value) === 'getfeatureinfo') {
-                        $url = $instance->getSource()->getGetFeatureInfo()->getHttpGet();
-                    } elseif (strtolower($value) === 'getlegendgraphic') {
-                        $url = $instance->getSource()->getGetLegendGraphic()->getHttpGet();
-                    }
-                    break;
-                }
+        $requestType = null;
+        foreach ($getParams as $key => $value) {
```
sveta-wl: Maybe it's better to do it like this:
`$request = $this->getRequest();
$requestType = $request->query->get('request')` ?

AxxL:
> For extreme hazard testing, please use the POC branch based on this one:
https://github.com/mapbender/mapbender/tree/fix/tunnel-getlegendgraphic-POC-pull616

So, what's your choice?

src/Mapbender/WmsBundle/Element/WmsLoader.php
```diff
@@ -247,7 +247,7 @@ protected function getInstances()
                 $instances[] = $instConfig;
             }
         }
-        return new Response(json_encode(array("success" => json_encode($instances))), 200, array(
+        return new Response(json_encode(array("success" => $instances)), 200, array(
```
sveta-wl: Don't you mind to use here JsonResponse:
`return new JsonResponse(['success' => $instances]);` ? It takes a response status code as second argument, by default it's 200.

src/Mapbender/CoreBundle/Controller/ApplicationController.php
werrolf: This is WMS-/OGC-specific quirk where query parameter names are case-insensitive. I.e. some people use REQUEST=GetMap, some people use request=GetMap. If I do request->query->get('request') I will not catch the upper-case variant.

I myself would prefer using a case-insensitive RequestParameterBag implementation to make code like this more readable, and to have this behaviour documented in a central place.

src/Mapbender/WmsBundle/Element/WmsLoader.php
werrolf: Yeah, we can do this. I tried to do only minimal modification here, but I changed the line anyway so might as well do it right.

werrolf:
@AxxL if there are no open issues, this pull will be merged, the POC-pull616 will be deleted. I don't want the change in POC to be merged.

The branch in this pull does not strip information from the DB. It will only affect how newly created / updated WMS instances are stored. They will be more compact.

The POC branch is for demonstration purposes, that the WMS stuff will continue to work even if the layer information currently stored in the DB is forcibly stripped from all objects. That's basically a quick simulation what will happen to the DB, on the non-POC branch, after deleting and recreating all WMS instances.

tl;dr: POC branch was a demo / testing helper and will be deleted. The non-POC branch is the merge candidate.

werrolf:
@sveta-wl I changed the WmsLoader to use JsonResponse, I tested and it works the same as before.

PR #645, opened 2017-10-13
Hotfix/custom doctrine param converterdefintion

LazerTiberius:
Some custom UserProvider used this class and broke with the 3.6.x update.

### DIFF
--- a/src/Mapbender/Component/CustomDoctrineParamConverter.php
+++ b/src/Mapbender/Component/CustomDoctrineParamConverter.php
@@ -36,17 +36,17 @@ class CustomDoctrineParamConverter implements ParamConverterInterface
 
     /**
      * @param Request        $request
-     * @param ParamConverter $configuration
+     * @param ParamConverter $paramConverter
      * @return bool|void
      */
-    public function apply(Request $request, ParamConverter $configuration)
+    public function apply(Request $request, ParamConverter $paramConverter)
     {
-        $this->configuration = $configuration;
-        $class               = $configuration->getClass();
-        $options             = $this->getOptions($configuration);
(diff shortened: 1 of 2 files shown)


src/Mapbender/Component/CustomDoctrineParamConverter.php
```diff
@@ -28,16 +38,21 @@ public function __construct(\Doctrine\Bundle\DoctrineBundle\Registry $registry =
         $this->registry = $registry;
```
werrolf: Returning from a constructor is rarely useful. What if it is null and you return? Then the instance attribute is not set to null. But it already is null, even without the explicit default. So the check doesn't change anything. Because of the type declaration, it can only ever be null or an object anyway.

The real issue here is that the object loses most of its function without the registry. If there's a use case for that, we could do a SupportsAbsolutelyNothingParamConverter that has a working apply method but can't do anything else.

LazerTiberius:
Actually, the interface defines ParamConverter type as argument, so using  the Interface as type will not work

src/Mapbender/Component/CustomDoctrineParamConverter.php
```diff
@@ -1,87 +1,87 @@
...
-
-    protected $configuration = null;
-
-    public function __construct(\Doctrine\Bundle\DoctrineBundle\Registry $registry = NULL)
+ */
+class CustomDoctrineParamConverter implements ParamConverterInterface
+{
+
+    protected $configuration;
+    protected $registry = null;
```
eSlider: Don't neet to set to `null` like `$configure`

PR #648, opened 2017-10-18
Fix/tunnel getlegendgraphic 306

werrolf:
This is basically a dupe of #616 for the release/3.0.6 branch.
https://github.com/mapbender/mapbender/pull/616

Had to do a new pull because protected and all that.
There were some conflicts in WmsInstanceLayerEntityHandler because of some variable hoisting / refactoring in there.

@AxxL I duped the changelog entry because technically we broke the timeline. Remember, we wanted to have this in the 3.0.5 series, so this is a special occasion. Please comment if you're ok with having the CHANGELOG.md entry twice.

If you compare the commit hashes you'll see that they are all the same, only the final two (merge) commits are newish.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+* **v.3.0.6.4**
+    - Fix GetLegendGraphic tunnel
+
 * **v.3.0.6.3**
     - Fix WMS with Scale fails to load #584 - see commit message #2783540 for more information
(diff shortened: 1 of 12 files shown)


werrolf:
If there are no comments I will super-cow-merge tomorrow. The change has already been reviewed, tested and approved on the original release/3.0.5 branch base.

werrolf:
Hello it's happening thx

Issue #639, opened 2017-09-28
Element grandchildren require reimplementation of getType, getFormTemplate etc


werrolf:
The Element base class was refactored to implement certain methods magically, and "calculate" form templates, concrete AdminType subclass and other bits via late static binding. This is generally a good idea because it removes the need for error-prone copy&paste in direct child classes.

For grandchildren of Element however (children of a concrete Element subclass), the current implementation prevents immediate inheritance of these calculated values. This means if you want to inherit the AdminType, form template etc from a concrete parent element, the relevant getter methods must again be reimplemented.

What we want is a mechanism that minimizes the need for copy&paste in all cases:
1. direct inheritance from abstract Element where you **want** your own AdminType, form template etc
2. second-gen inheritance from concrete Element child where you **want to inherit** AdminType, form template etc
3. second-gen inheritance from concrete Element child where you want to **override** AdminType, form template etc

The change to late static binding and (desirable!) removal of c&p methods leads to problems in older applications upgrading to Mapbender 3.0.5 or later.

werrolf:
Demo case:
```php
<?php


namespace Mapbender\CoreBundle\Command;


use Mapbender\CoreBundle\Component\Element;
use Mapbender\CoreBundle\Element\HTMLElement;
use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;


/**
 * Direct child class of Element, should by default get its own distinct template name, admin type etc
 * (and it does!)
 */
class FirstGenElement extends Element
{
}

/**
 * Child of an already defined concrete Element child class (>= grandchild of Element)
 * We need a method here to inherit template, admin type etc from parent.
 *
 * HTMLElement was chosen because it has no own implementation of "getType" and "getFormTemplate" methods. It
 * was refactored in commit 663e1c2eb796ad55392880de6062cb9a1be0ec5b to inherit these methods from the parent
 * implementation in Mapbender\CoreBundle\Component\Element
 */
class SecondGenButton extends HTMLElement
{
}

class Omg639Command extends ContainerAwareCommand
{
    public function configure()
    {
        $this->setName('omg639');
    }

    public function execute(InputInterface $input, OutputInterface $output)
    {
        echo "### First-gen Element: form template / admin type:\n";
        echo " " . FirstGenElement::getFormTemplate() . "\n";
[...]

PR #649, opened 2017-10-20
Fix/element inheritance 639 [restore BC]

werrolf:
See #639 
This may seem like a lot of code, because it is.

This addresses a BC break somewhere in the 3.0.5 timeline where Elements inheriting from each other would at some point require `getType`, `getFormTemplate`, potentially `render` re-implementations, because automagically calculated template paths and admin types would be different from the earlier (always explicit) values coded into each element.

The motivation for that refactoring was to get rid of endless c&p of the three methods mentioned above.

This pull kind of "concludes" that work. C&P is not necessary for new elements inheriting directly from Element, and no necessity exists to override methods in child classes of concrete elements either, if the (intuitive) inheritance of AdminType and templates is desired. Old-style overrides (just reimplementing any or all of the methods) still works on this branch. Only the automatically generated values are affected at all.

Because this was already a BC break, and this pull is supposed to reestablish BC, its merge target is 3.0.5.

One issue was that there has not been a method to get the template path for the frontend. This logic was so far hard-embedded into the render method(s), and I assume a lot of C&Ping of that render method has been going on for that reason alone that makes it hard to analyze.
To cover all potential cases, I included some fine-grained configurability of which of the automagically calculated values exactly should be inherited from the parent element.
This should support all sorts of mixed cases (inherit automatic AdminType but not automatic frontend template path, inherit frontend template but not admin type etc).

If there are no such complex use cases this configurability could be stripped back to make for a more concise pull.

To see exactly what's happening, the (not included) Omg639Command code from the original issue #639 may be a good starting point.

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Element.php
+++ b/src/Mapbender/CoreBundle/Component/Element.php
@@ -658,7 +658,10 @@ abstract class Element
      *
      * By default, inheritance is enabled.
      * If you want to control this inheritance fine-grained, see the static $inheritedFromParent attribute.
-     * If you just want to replace one of the values, implement the getType
+     * If you just want to replace one of the values, reimplement the getType, getFormTemplate, getHtmlTemplatePath
+     * methods.
+     * If you want your values to be automatically calculated but NOT inherited from the parent, adjust the
+     * value of the $inheritedFromParent static property accordingly.
      *
      * @param string $what see above for valid values
      * @return bool
--- a/src/Mapbender/CoreBundle/Component/Element.php
+++ b/src/Mapbender/CoreBundle/Component/Element.php
@@ -58,6 +58,34 @@ abstract class Element
     protected static $title = "mb.core.element.class.title";
 
     /**
+     * @var string[]
+     * Determines if "frontendTemplate", "formTemplate", "adminType" should be inherited from the parent class
+     * EVEN IF THEY ARE AUTOMAGICALLY CALCULATED. The three strings above are the valid / effective values.
+     * Also supported is "all".
+     *
+     * E.g. a SuperHTMLElement extends HTMLElement extends Element, with all inheritance left enabled (default)
+     * and no own implementations would return
+     *  getHtmlTemplatePath(false) => "MapbenderCoreBundle:Element:htmlelement.html.twig"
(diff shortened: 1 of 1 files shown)


src/Mapbender/CoreBundle/Component/Element.php
```diff
@@ -368,10 +368,10 @@ public function trans($key, array $parameters = array(), $domain = null, $locale
...
 
     /**
-     * Get the element configuration form type.
+     * Get the element configuration form type. By default, this class needs to have the same name
+     * as the element, suffixed with "AdminType" and located in the Element\Type sub-namespace.
      *
-     * Override this method to provide a custom configuration form instead of
-     * the default YAML form.
+     * Override this method and return null to get a simple YAML entry form.
      *
```
werrolf: For the change in behavior redocumented here, see https://github.com/mapbender/mapbender/commit/387adf1e90e733249f436ec67ad11daf05bc4249#diff-aa19882f5b611faae7e41b317a61faf7L323
Could be another BC break.

src/Mapbender/CoreBundle/Component/Element.php
```diff
@@ -57,6 +57,34 @@
...
+     *  getType()                  => "Mapbender/CoreBundle/Element/Type/HTMLElementAdminType"
+     *
+     * With inheritance off (empty array or or other falsy value, it would instead return:
+     *  getHtmlTemplatePath(false) => "MapbenderCoreBundle:Element:super_htmlelement.html.twig"
+     *  getHtmlTemplatePath(true)  => "MapbenderCoreBundle:ElementAdmin:super_htmlelement.html.twig"
+     *  (same:) getFormTemplate()  => "MapbenderCoreBundle:ElementAdmin:super_htmlelement.html.twig"
+     *  getType()                  => "Mapbender/CoreBundle/Element/Type/SuperHTMLElementAdminType"
+     * I.e. with inheritance off, new templates and new admin type would be required by default.
+     *
+     * @todo: add widgetName to inheritables? Does that make sense?
```
eSlider: Yes. There is no element without a widget fronted. Each widget has a unique name this need to be associated with an element PHP class. Potentially: more than one element back end can have one front-end widget.

werrolf:
Superseded by [a better, less buggy implementation](https://github.com/mapbender/mapbender/pull/743), this pull will not be merged and remain closed.

PR #655, opened 2017-11-13
Change removed function call  Mapbeder.Dimension.getValue() to new fu…

LazerTiberius:
Dimension Handler is not working on some projects. Hit me up for further information and internal tickets.


AxxL:
@LazerTiberius @CharlyUnicorn 

You had some project-specific ticket, can you create a public one, describing in a short way what went wrong?

CharlyUnicorn:
Hi @AxxL 
I think, that @LazerTiberius and @atd-schubert  can tell you more (they are currently working on it), but here some facts: 
* a wms with time dimension was in the application
* the dimensions handler could be activated by a checkbox in the context menü of the layertree (then you could slide through the years) 
* after update of the mapbender to the newest version -> the checkbox could be activated once and the slider worked once. After that, the function could no longer be used and freezed (no action when changing the slider position )

Issue #595, opened 2017-08-09
[Scale Selector] No refresh on slow WMS services


AxxL:
Steps to reproduce:

* Get default application
* New Layer, choose something complex and "slow" like this one here: https://wms2016.wheregroup.com/cgi-bin/alkis/qgis_mapserv.fcgi?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities
* Start Extent (minx, miny, maxx, maxy):
** 348367.778537
** 5526659.925436
** 356331.752799
** 5535311.817739
** SRS: EPSG:25832

Zoom around.

Observed: Switching the Scale Selector or Zooming with the Navigation Toolbar doesn't update the Scale Selector itself everytime. This is not a regular case. AFAIK it is best visible if the service needs some time to render.

So if you're in 1:500 it may still show 1:10000. You can double check with the Navigation bar and the hidden scale-display behind the sidebar.

Expected result: The Scale Selector shows the actual state.

PR #657, opened 2017-11-15
Fix displaying scale value in scale selector

eSlider:
* Fix displaying scale value in scale selector
* Clean scaleselector element JS code

Closes: #647
Associated: #595

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.scaleselector.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.scaleselector.js
@@ -1,65 +1,152 @@
 (function($) {
 
     $.widget("mapbender.mbScaleSelector", {
-        options:      {
+
+        options: {
+            /**
+             * Target widget id string
+             */
             target: null
         },
-        elementUrl:   null,
-        _create:      function() {
-            var widget = this;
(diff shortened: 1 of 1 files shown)


CharlyUnicorn:
The scale display on the scale selector and the fix displaying in the bar on the right and in the map respond correctly. Nice work. 
Tested with two different ALKIS-WMS.

auswahl_251




The only problem/change request I noticed is the display of the scale in the map. The element is probably rounded and therefore in the example screen shot instead of the 1: 100500 a 1: 101K appears, which could possibly confuse some users. Here you could think about a configuration possibility in the administration, which allows the selection of the rounding and display behavior.

astroidex:
Hi @eSlider hi @CharlyUnicorn thanks for the fix and the testing. I will have a look at it next week.

eSlider:
@CharlyUnicorn,  101K isn't a bug RTM: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Math/round

eSlider:
@CharlyUnicorn, one more, the bug I fixed depends on scale selector behavior, not on scale display element.

eSlider:
@werrolf ping.

werrolf:
Huh?

PR #658, opened 2017-11-15
Fix displaying scale value in scale selector

eSlider:
* Fix displaying scale value in scale selector
* Clean scaleselector element JS code

Closes: #647
Associated: #595


src/Mapbender/CoreBundle/Resources/public/mapbender.element.scaleselector.js
```diff
@@ -1,65 +1,152 @@
...
+                var option = $("<option/>");
+                var formattedValue = Math.round(value);
+
+                option.attr('value', formattedValue).html(formattedValue);
+                select.append(option);
             });
 
             select.change($.proxy(widget._zoomToScale, widget));
-            $('option[value="' + scale + '"]', select).attr('selected', true);
+            select.val(scale);
```
AxxL: You set the value of the select with "scale" which come from olMap (line 58) But the options from select are from the mbMap scales. (line 61 ff.). Why not consistent?

eSlider: @AxxL, as you said options from select are "options from select". `scale` - is current value from OL Map which is set by own element configuration. `options` has any values. Select has this `options` and need current `scale` value to show. `olMap` - is an object to be manipulated by select. mbMap.scales() - are option values from map element configuration. What is wrong?

Issue #690, opened 2017-11-23
PrintClient feature.data print does not handle Umlaute öüä utf8_decode is needed


astroidex:

astroidex:
* very small issue
* see pull request https://github.com/mapbender/mapbender/pull/691

AxxL:
We cannot reproduce this error. What happens?

astroidex:
* see also ticket for the new feature https://github.com/mapbender/mapbender/issues/695
* see also documentation for the new feature 

https://github.com/mapbender/mapbender-documentation/blob/feature/printclient-feature-data-print/en/bundles/Mapbender/CoreBundle/elements/printclient.rst
description starts at row: 229

astroidex:
* pull requests merged #691 #701

PR #691, opened 2017-11-23
PrintClient - fixed feature.data print utf8_decode

astroidex:
* fixed issue #690 PrintClient feature.data print does not handle Umlaute öüä utf8_decode is needed
* added utf8_decode

### DIFF
--- a/src/Mapbender/PrintBundle/Component/PrintService.php
+++ b/src/Mapbender/PrintBundle/Component/PrintService.php
@@ -462,7 +462,7 @@ class PrintService
                             $attribute = substr(strrchr($k, "."), 1);
                             $pdf->MultiCell($this->conf['fields'][$k]['width'],
                                 $this->conf['fields'][$k]['height'],
-                                $feature->getAttribute($attribute));
+                                utf8_decode($feature->getAttribute($attribute)));
                         }
                         break;
                 }


AxxL:
@astroidex: Associated Ticket? Steps to reproduce? Testscenario?

@swinkelmann: I've adjusted the Milestone, Label. I put you on Assignee.

AxxL:
@werrolf Please review the PR in our 4-eye-process.

astroidex:
This error occurs when you use print client to print feature data via feature.column (you can do this via digitizer or feature info for example). The code missed to utf8 decode the text. utf decode is done for all other textes.
@werrolf your ok is missing - I can show you if you do not have a running configuration and you have questions

astroidex:
As I mentioned. I would like to show it to @werrolf and @AxxL and then you will understand. Sorry. I did not have the time yet to show you. Would be nice not to block a PR if the there was not the time to explain. 

Only change is: utf8_decode($feature->getAttribute($attribute)));

And I will explain what happens. Have time next week.

astroidex:
Here is the information for you to reproduce the error:
* see also ticket for the new feature https://github.com/mapbender/mapbender/issues/695
* see also documentation for the new feature https://github.com/mapbender/mapbender-documentation/blob/feature/printclient-feature-data-print/en/bundles/Mapbender/CoreBundle/elements/printclient.rst
description starts at row: 229
* I also attach screenshots to show the problem

printclient-utf8-encoding-without-utf8_decode

printclient-utf8-encoding-with-utf8_decode

If you don't min I would show you the functionality and problem next week. It is not so easy to configure and is much faster when we do it together.

werrolf:
Fix included in already merged pull #701

PR #699, opened 2017-12-11
Change visibility of some fields in WmsLayerSource and Wmc Entites

LazerTiberius:
Some fields in both entities had the visibility set to protected as a workaround for some prehistoric doctrine bug. But with these visibilities the entities are invalid and doctrine:schema:validate would fail and you could not create a new db with the console commands. The doctrine bug is now fixed, so we can change the visibility to the correct protected value.


werrolf:
I stripped back the added blank lines in the WmsLayerSource entity separating the annotations from the property declarations. Stylistic extra blanks should go above comments / annotations, not below.

Issue #694, opened 2017-12-06
Source instance delete confirmation broken (locale en)


werrolf:
Commit a9834b441f6487ea6fc34f1d8a913a1e30189fa2 changed the translation key ```mb.manager.admin.instance.delete.title``` to just ```mb.manager.admin.instance.delete``` (dropping the .title) in form-layersets.html.twig, some translation yamls were updated, ignoring that there also was a ```mb.manager.admin.instance.delete.confirm``` translation key used for the dialog that pops up when you click the cross icon. This key is no longer defined.

Now the confirmation dialog says, e.g.
mb.manager.admin.instance.delete WMS BB ALKIS?
The hover tooltip for the cross icon is the same, too in locale en.

We *need* a different text for the body of the confirmation box.
Revert the folding of both translations onto a single key.

Assigning to the person who commited the change.

astroidex:
* please note that benno worked on the translation and fixed it for all bundles. The fixed translation is in branch **hotfix/translations**  and is waiting for being merged in the next version.

https://github.com/mapbender/mapbender/tree/hotfix/translations

astroidex:
* translation was done in summer (Juli 2017) during the Translation Sprint at FOSS4G Europe in Paris - see email https://lists.osgeo.org/pipermail/mapbender_users/2017-July/005259.html
* You find the translation ticket here https://github.com/mapbender/mapbender/issues/557
* there is no pull request yet for hotfix/translation as I did not know to which branch  to make the pull request 
* please see internal Ticket chili 8345 for more information

eSlider:
@werrolf you are wrong. I added the `title`.

eSlider:
@werrolf, please review https://github.com/mapbender/mapbender/pull/696

eSlider:
@werrolf, depence on #497

werrolf:
Resolved in pull #700

PR #497, opened 2017-03-30
fixed typos and changed some words

astroidex:
* see ticket https://github.com/mapbender/mapbender/issues/495
* fixed some typos and used button instead of word tabs in sidepane

PR #700, opened 2017-12-11
Integrate community-contributed translations from hotfix/translations

werrolf:
Updates translations, replaces English / dummy placeholders in French, Spanish, Russian and Italian locales with community-contributed proper translations, adds Turkish translation.

This branch changes only messages.*.yml files and src/Mapbender/ManagerBundle/Resources/views/Application/form-layersets.html.twig.

Bundles fix for issue #694 (instance.delete vs instance.delete.confirm/  instance.delete.title translation keys => the twig change).
Bundles substitute for pull #497, fixing the capitalization for "Delete instance" without structure breaks, discarding the other changes in that pull.

This branch was apparently based on feature/ol-cesium at d11dd2fd1bde139225a388ddb6d125cb24562260, which has already been merged to release/3.0.6, but reverted later. This shouldn't cause further issues.

Need this in 3.0.6 because many locales are currently filled with placeholder content, which is worse than not having them included at all.


werrolf:
There are still some obvious placeholders in there.
For everything with "... [Placeholder]" or "__mb....."  we're better off if we just remove them; then they'll be sourced from the fallback locale. This may not be what you want for the actual translation creation workflow, but that's what we should aim for in the repository state.

Trying to introspect this thing to produce a full list of placeholder / non-localized strings ...

Issue #692, opened 2017-11-23
PrintClient with feature.data - print runs on an error if the feature.data column does not exist


astroidex:
* if you define a feature.data definition in your print template f.e. feature.hauptcode that does not exist in the feature type, the print exits the pdf generation with an error
* in this case feature.hauptcode is written with a k feature.hauptkode

`CRITICAL - Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "Notice: Undefined index: hauptcode" at /data/git/mainz/application/vendor/mapbender/data-source/Entity/DataItem.php line 158 `

ToDo: Ignore is feature.column is not present in the feature type

astroidex:
The errors in vendor/mapbender/data-source/Entity/DataItem.php at line 159  -

```
Notice: Undefined index: abstractNIX
500 Internal Server Error - ContextErrorException 
```

FIX could be error_reporting(0);

```
    public function getAttribute($name){
        $attributes = $this->getAttributes();
        error_reporting(0);
        return $attributes[$name];
    }
```

astroidex:
* fixed with pull reuqest #701

PR #701, opened 2017-12-15
PrintClient - fixed feature exists

astroidex:
* if the template with feature is defined and the template is used in a use case without an feature an error occurs
* the pull request checks this and prevents the error
* in combination with https://github.com/mapbender/mapbender/pull/691
* see ticket #692

### DIFF
--- a/src/Mapbender/PrintBundle/Component/PrintService.php
+++ b/src/Mapbender/PrintBundle/Component/PrintService.php
@@ -456,13 +456,13 @@ class PrintService
 
                         // fill digitizer feature fields
                         if(preg_match("/^feature./", $k)){
-                            if($feature == false){
+                            if(!isset($feature) OR $feature == false){
                                 continue;
                             }
                             $attribute = substr(strrchr($k, "."), 1);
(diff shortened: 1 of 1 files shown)


src/Mapbender/PrintBundle/Component/PrintService.php
```diff
@@ -456,13 +456,13 @@ private function buildPdf()
...
                         if(preg_match("/^feature./", $k)){
-                            if($feature == false){
+                            if(!isset($feature) OR $feature == false){
                                 continue;
                             }
                             $attribute = substr(strrchr($k, "."), 1);
                             $pdf->MultiCell($this->conf['fields'][$k]['width'],
                                 $this->conf['fields'][$k]['height'],
-                                $feature->getAttribute($attribute));
+                                utf8_decode($feature->getAttribute($attribute)));
```
eSlider: @astroidex, to save decoded `$attrubute` you need to assign new value into it:
`$attribute =  utf8_decode($feature->getAttribute($attribute));`

Do it's rightly worked before? 
Other thing is: one `)` sign more, at eol, as expected.

Issue #554, opened 2017-07-05
New Option in Layerset Instances "Layer order reversed"


AxxL:
With the rising emergence of the fabulous QGIS Server we need to add an option how WMS layers are put into the layertree and in which order they are rendered. Right now we followed the mapserver approach, where we parsed the capabilities from top to bottom. First layer was drawn first, second layer at second on top of the first and so on. With QGIS Server the Capabilties document follows the configuration of the QGIS project and the desktop layer order of the layertree. So in general points are at top and polygons at the bottom.

So, place an option into the layerset dialog (when adding a wms to an application) where you set:

- [x] Layerorder reversed (QGIS style)

For backwards compatibility: Set it to false for default.

That means: If the option is not checked, the layers are placed into the layertree like it is, beginning at the bottom and ending at the top. The service is displayed in the layertree in reversed order.

If checked, layers are placed into the layertree from top to bottom, the layertree shows the service not in reversed order but like the definition in the Capablities

``` org
|                                        | layer order not reversed | layer order reversed |
|----------------------------------------+----------------------+--------------------------|
| WMS Capabilities (from top to bottom)  | polygon, line, point | point, line, polygon     |
| Layerset-Instance (from top to bottom) | polygon, line, point | point, line, polygon     |
| Layertree  (from top to bottom)        | point, line, polygon | point, line, polygon     |
```

Why do we know what is the default? See the OGC WMS 1.3 Demo and Reference Server documentation: http://www.opengeospatial.org/standards/wms/quickstart

WMS-URL: http://metaspatial.net/cgi-bin/ogc-wms.xml?REQUEST=GetCapabilities&SERVICE=WMS&VERSION=1.3

AxxL:
A new demo-WMS:

https://wms2016.wheregroup.com/cgi-bin/schwalmtalgroup/qgis_mapserv.fcgi?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities

Layers:

- 1Punkt
  - 11Bohrungen
- 2Linie
  - 22Hoehenlinien
- 3Polygon
  - 31Bodenarten (not visible per default)
  - 32Landnutzung (this is the visible one per default)

AxxL:
New demo app: https://test.mapbender.org/mapbender/application/test_554

AxxL:
Example for OGC Reference WMS (traditional style): https://test.mapbender.org/mapbender/application/test_554_b
Is in Britain.

AxxL:
[...]

PR #704, opened 2017-12-19
Feature/reversible layer order

werrolf:
To solve issue #554 

**THIS REQUIRES A DATABASE SCHEMA UPDATE** because it adds a column to mb_wms_wmsinstance.
```$ app/console doctrine:schema:update --force```

(use the same command if / after you leave this feature branch, then the new column will be dropped again)

Pull adds layer rendering order reversal, controllable by the admin on a per-layerset instance basis.
Default behavior is the same as before. 

There's a new dropdown in the backend (via application => edit => Layersets tab => add or edit instance).
Dropdown label and value labels are translated for DE and EN locales, see WmsBundle/Resources/translations/messages.*.yml.

The default layer order for existing instances is fixed to the old behavior, so running installations can be upgraded with no change on ordering for existing layersets.
The default layer order for **new** instances can be configured in parameters.yml (not included here, but see below).

Open Tasks: Changelog, User Documentation, parameters.yml.dist extension (external; goes into mapbender-starter)

For parameters.yml.dist, I have the following suggestion (tested)
```
    # default layer order when creating *new* WMS layerset instances
    # allowed values are either
    # * "standard": top-down rendering in GetCapabilities order; also the default if this parameter is not defined
    # * "reverse": bottom-up, for QGIS server, ArcGIS etc
    wms.default_layer_order: standard
```

### DIFF
--- a/src/Mapbender/WmsBundle/Component/WmsInstanceConfigurationOptions.php
+++ b/src/Mapbender/WmsBundle/Component/WmsInstanceConfigurationOptions.php
@@ -396,29 +396,29 @@ class WmsInstanceConfigurationOptions extends InstanceConfigurationOptions
                 }
             }
         }
-        $vendorsecifics = array();
+        $vendorspecifics = array();
         foreach ($entity->getVendorspecifics() as $key => $vendorspec) {
             $handler = new VendorSpecificHandler($vendorspec);
             /* add to url only simple vendor specific with valid default value */
             if ($vendorspec->getVstype() === VendorSpecific::TYPE_VS_SIMPLE && $handler->isVendorSpecificValueValid()) {
-                $vendorsecifics[] = $handler->getConfiguration();
+                $vendorspecifics[] = $handler->getConfiguration();
                 $help             = $handler->getKvpConfiguration(null);
                 $options->setUrl(UrlUtil::validateUrl($options->getUrl(), $help, array()));
             }
         }
 
         $bboxes = $entity->getRootLayer()->getSourceItem()->getMergedBboxes();
-        // reformat bboxes to coordinate arrays keyed on SRS
-        $bboxesOut = array();
+        // reformat bboxes to coordinate arrays, keyed on SRS
+        // @todo: when the same SRS comes up again, calculate the minimum box (current: last occurence per SRS wins)
+        $bboxArrays = array();
         foreach ($bboxes as $bbox) {
-            $bboxesOut[$bbox->getSrs()] = array(
+            $bboxArrays[$bbox->getSrs()] = array(
                 floatval($bbox->getMinx()),
                 floatval($bbox->getMiny()),
                 floatval($bbox->getMaxx()),
                 floatval($bbox->getMaxy()),
             );
         }
-
 
         $options->setProxy($entity->getProxy())
             ->setVisible($entity->getVisible())
@@ -427,11 +427,11 @@ class WmsInstanceConfigurationOptions extends InstanceConfigurationOptions
             ->setTransparency($entity->getTransparency())
             ->setOpacity($entity->getOpacity() / 100)
             ->setTiled($entity->getTiled())
-            ->setBbox($bboxesOut)
+            ->setBbox($bboxArrays)
             ->setDimensions($dimensions)
             ->setBuffer($entity->getBuffer())
             ->setRatio($entity->getRatio())
-            ->setVendorspecifics($vendorsecifics)
+            ->setVendorspecifics($vendorspecifics)
             ->setVersion($entity->getSource()->getVersion())
             ->setExceptionformat($entity->getExceptionformat());
 
--- a/src/Mapbender/WmsBundle/Component/WmsInstanceConfigurationOptions.php
+++ b/src/Mapbender/WmsBundle/Component/WmsInstanceConfigurationOptions.php
@@ -68,6 +68,9 @@ class WmsInstanceConfigurationOptions extends InstanceConfigurationOptions
      * ORM\Column(type="decimal", scale=2, options={"default" = 1.25})
      */
     public $ratio = 1.25;
+
+    /** @var  string */
+    public $layerOrder;
     
     /**
      * Returns a version
@@ -298,6 +301,24 @@ class WmsInstanceConfigurationOptions extends InstanceConfigurationOptions
     }
 
     /**
+     * @return string
+     */
+    public function getLayerOrder()
+    {
+        return $this->layerOrder;
+    }
+
+    /**
+     * @param string $layerOrder for valid values see WmsInstance constants LAYER_ORDER_...
+     * @return $this
+     */
+    public function setLayerOrder($layerOrder)
+    {
+        $this->layerOrder = $layerOrder;
+        return $this;
+    }
+
+    /**
      * @inheritdoc
      */
     public function toArray()
@@ -318,6 +339,7 @@ class WmsInstanceConfigurationOptions extends InstanceConfigurationOptions
             "dimensions" => $this->dimensions,
             "buffer" => $this->buffer,
             "ratio" => $this->ratio,
+            "layerOrder" => $this->layerOrder,
         );
     }
 
@@ -370,6 +392,9 @@ class WmsInstanceConfigurationOptions extends InstanceConfigurationOptions
             }
             if (isset($options["exception_format"])) {
                 $ico->exceptionformat = $options["exception_format"];
+            }
+            if (isset($options["layerOrder"])) {
+                $ico->setLayerOrder($options["layerOrder"]);
             }
         }
         return $ico;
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Comment by werrolf** [comment]
@AxxL please test and give feedback

**Comment by werrolf** [comment]
I noticed some unexpected links between WmsInstance (which is undergoing a slight behavioural change with this) and WmcLoader. I have not tested interactions with WmcLoader myself yet.

We should also consider extending WmsLoader with similar functionality.

**Comment by AxxL** [comment]
@werrolf I've got an issue with the glasses symbol in the layertree. Sandwich symbol of the service, Glasses icon. The map should zoom to the given extent of the service.

That doesn't work right now. Without the change (other branch) it still works.

**Comment by werrolf** [comment]
Fixed just now, please try again.

File: src/Mapbender/WmsBundle/Component/WmsInstanceEntityHandler.php [hunk 1]
```diff
@@ -143,12 +143,34 @@ public function create()
...
     }
 
+    /**
+     * Populates the layer order of an (assumed) newly created WmsInstance
+     *
+     * @param WmsInstance $entity
+     */
+    public function initializeLayerOrder(WmsInstance $entity)
+    {
+        $layerOrderDefaultKey = 'wms.default_layer_order';
```
eSlider: Maybe better with `public static const`?

src/Mapbender/WmsBundle/Component/WmsInstanceConfigurationOptions.php
```diff
@@ -369,7 +393,75 @@ public static function fromArray($options)
...
+     *
+     * @param WmsInstance $entity
+     * @return static
+     */
+    public static function fromEntity(WmsInstance $entity)
+    {
+        $options = new static();
+
+        $options->setUrl($entity->getSource()->getGetMap()->getHttpGet());
+        $dimensions = array();
```
eSlider: The function becomes really big. `getActiveDimensionConfigurations()` at this point?

src/Mapbender/WmsBundle/Component/WmsInstanceLayerEntityHandler.php
```diff
@@ -185,13 +185,12 @@ public function generateConfiguration()
     {
         $configuration = array();
         if ($this->entity->getActive() === true) {
-            $children = null;
+            $children = array();
             if ($this->entity->getSublayer()->count() > 0) {
-                $children = array();
                 foreach ($this->entity->getSublayer() as $sublayer) {
-                    $instLayHandler = self::createHandler($this->container, $sublayer);
+                    $instLayHandler = new WmsInstanceLayerEntityHandler($this->container, $sublayer);
```
eSlider: Much better! :+1:

src/Mapbender/WmsBundle/Controller/RepositoryController.php
```diff
@@ -221,18 +225,18 @@ public function updateformAction($sourceId)
     public function updateAction($sourceId)
     {
         $request         = $this->get('request');
-        $source          = $this->getDoctrine()->getRepository("MapbenderCoreBundle:Source")->find($sourceId);
+        $source          = $this->loadEntityByPk("MapbenderCoreBundle:Source", $sourceId);
```
eSlider: Maybe better to create and use `getSourceById($sourceId)` method?

src/Mapbender/WmsBundle/Controller/RepositoryController.php
```diff
@@ -345,11 +346,11 @@ public function deleteAction($sourceId)
         $aclProvider->deleteAcl($oid);
 
         foreach ($wmsinstances as $wmsinstance) {
-            $wmsinsthandler = EntityHandler::createHandler($this->container, $wmsinstance);
+            $wmsinsthandler = new WmsInstanceEntityHandler($this->container, $wmsinstance);
```
eSlider: :+1:

src/Mapbender/WmsBundle/Controller/RepositoryController.php
```diff
@@ -496,9 +495,7 @@ public function instanceLayerPriorityAction($slug, $instanceId, $instLayerId)
             $em->persist($inst);
         }
         $em->flush();
-        $wmsinstance = $this->getDoctrine()
-            ->getRepository("MapbenderCoreBundle:SourceInstance")
-            ->find($instanceId);
+        $wmsinstance = $this->loadEntityByPk("MapbenderCoreBundle:SourceInstance", $instanceId);
```
eSlider: Camelcase?

src/Mapbender/WmsBundle/Controller/RepositoryController.php
```diff
@@ -530,7 +525,7 @@ public function instanceEnabledAction($slug, $instanceId)
             );
         } else {
             $enabled_before = $wmsinstance->getEnabled();
-            $enabled        = $enabled === "true" ? true : false;
+            $enabled        = $enabled === "true";
```
eSlider: :+1:

src/Mapbender/WmsBundle/Entity/WmsInstance.php
```diff
@@ -136,7 +147,7 @@ public function getId()
     /**
      * Returns dimensions
      *
-     * @return array of DimensionIst
+     * @return DimensionInst[]
```
eSlider: :+1:

src/Mapbender/WmsBundle/Entity/WmsInstanceLayer.php
```diff
@@ -566,4 +566,13 @@ public function __toString()
     {
         return (string) $this->getId();
     }
+
+    /**
+     * @return boolean
+     */
+    public function isRoot()
```
eSlider: :heart:

src/Mapbender/WmsBundle/Resources/translations/messages.de.yml
```diff
@@ -79,6 +79,9 @@ mb:
             cancel2: Abbrechen
             ratio: BBOX-Faktor
             buffer: Kachel-Puffer
+            layerOrder: Layer-Reihenfolge
+            layerOrder.standard: Standard (umgekehrt)
```
eSlider: Camelcase in translation YAML make the naming inconsistent. Maybe better  if:
```yaml
layer: 
  order:
    title: Layer-Reihenfolge
    standard: Standard (umgekehrt)
    reverse: QGIS Style (gleiche Reihenfolge)
```

src/Mapbender/WmsBundle/Controller/RepositoryController.php
werrolf: Yes. Kept variable naming as before to minimize diff, but camel-casing would be better.

src/Mapbender/WmsBundle/Controller/RepositoryController.php
werrolf: A simple (no arguments) ```->find()``` on the repository, as used here, actually uses doctrine metadata magic to auto-detect the primary key, which may not always be the id.

werrolf:
I'm kind of hesitant myself to go forward with the merge because of the schema update -.-
@AxxL I think this should be coordinated with a new feature release, so we can announce the required steps to upgrade. We should delay the merge as long as possible, ~"just in time" for the release. Agree?

astroidex:
Hi @werrolf, @AxxL . Great feature. I tested the branch. Works for me.

Just some suggestions:
1. checkbox instead of selectbox: what do you think about the idea of adding a checkbox instead of a selectbox (see screenshot)
mb_reverse
? 
pro:
- Would fit better to the actual setup of the administration 
- only one translation text needed in yml: 
en
 `reverselayerorder: reverse layer order`

de
 `reverselayerorder: umgekehrte Layerreihenfolge`

2. add explanation to documentation
* in the second step the explanation for the checkbox could move to the documentation. https://github.com/mapbender/mapbender-documentation/edit/release/3.0.6/en/book/quickstart.rst Line 286

3. Layer order
* as I know QGIS Server is not the only server using the different layer order (ArcGIS Server uses it too I read in a thread about QGIS layer order.)
* in WMS Spec it is not defined which is the right one

AxxL:
@werrolf Nice, works. Let's go. Documentation adjusted.

@astroidex 

> checkbox instead of selectbox

We had it considered, of course. There was a technical decision to do it as a drop-down. 

> add explanation to documentation

Rewrote the whole chapter: https://github.com/mapbender/mapbender-documentation/pull/30

> as I know QGIS Server is not the only server using the different layer order (ArcGIS Server uses it too I read in a thread about QGIS layer order.) in WMS Spec it is not defined which is the right one

It's part of the OGC WMS 1.3 Demo and Reference Server documentation: http://www.opengeospatial.org/standards/wms/quickstart and that WMS service. See description in https://github.com/mapbender/mapbender/issues/554 and documentation in  https://github.com/mapbender/mapbender-documentation/pull/30

PR #711, opened 2018-01-04
Feature/repeatable cli print

werrolf:
This is the ~minimal set of changes to make print jobs repeatable, so further clean ups can be (unit) tested and verified against fixed testing jobs.

This includes a new command mapbender:print:runJob which accepts a JSON or YAML. Job capturing is not included here, but JSON print jobs can, with some slight hacking and reformatting, be dumped from a Chrome inspector console or similar.

CLI support required changes to user checking (no Request object => no user) and Owsproxy interactions, where a controller action delegate from the CLI would throw a BadSignatureException. Because these parts needed to be changed anyway, they were also c&p folded.

C&P folding led to PrintService now inheriting from ImageExportService and very slight related cleanups.

This might sound scary, but the impact should be low. Almost all methods in both classes have historically been private, so customization via inheritance has never been possible to begin with.

This branch is lenient against a missing 'changeAxis' value to support jobs captured on 3.0.5.

### DIFF
--- a/src/Mapbender/PrintBundle/Component/ImageExportService.php
+++ b/src/Mapbender/PrintBundle/Component/ImageExportService.php
@@ -26,8 +26,8 @@ class ImageExportService
     protected $urlHostPath;
     /** @var array */
     protected $data;
-    /** @var string[] */
-    protected $requests = array();
+    /** @var string[] plain WMS URLs */
+    protected $mapRequests = array();
 
     public function __construct($container)
     {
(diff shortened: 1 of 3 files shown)


src/Mapbender/PrintBundle/Component/ImageExportService.php
```diff
@@ -144,8 +138,157 @@ private function getImages()
...
+                break;
+            case (preg_match("/image\/jpeg/", $contentType) ? $contentType : !$contentType) :
+                return imagecreatefromjpeg($storagePath);
+                break;
+            case (preg_match("/image\/gif/", $contentType) ? $contentType : !$contentType) :
+                return imagecreatefromgif($storagePath);
+                break;
+            default:
+                return null;
+                $this->getLogger()->debug("Unknown mimetype " . trim($response->headers->get('content-type')));
```
LazerTiberius: Does this work? How can the logger be called if the function returns before? Or is this some php black magic?

werrolf: It doesn't work. Same as before. It's an equivalent transform of old behaviour.
See this commit: https://github.com/mapbender/mapbender/pull/711/commits/3056d755b89a07314441a9ccc1d9e4e88f6b8111#diff-aa7793858e9f53e396ac560ee2b6a9b0L103

I have already worked on further improvements, including the reenabling of that logging message, but they are beyond the scope of this pull.

PR #712, opened 2018-01-09
Support non-GetLegendGraphic legends via instance tunnel

werrolf:
Instance tunnel is used when a service is secured (username + password) to hide auth details from client code.
Instance tunnel legend loopholes were closed in https://github.com/mapbender/mapbender/pull/616

This introduced a regression, losing non-GetLegendGraphic modes of retrieving a legend image for tunneled WMS. Namely, pulling a legend image from a layer's ```<Style>``` block in the capabilities document instead of the top-level ```<GetLegendGraphic>```.

This pull breaks out the logic for legend url selection into freestanding methods.

These methods are then reused as-is in the instance tunnel handling to get the same results for secured and unsecured WMS services.

This fixes the regression introduced in pull 616.

Collateral:
This pull also extends the ```<Style>``` legend picking to not always use the last Style node within a Layer, but instead it scans backwards inside the Layer until a Style node with a LegendUrl node is found.

This scanning change was implemented because we have found at least one (secured) WMS service where in the capabilites document the LegendUrl node is omitted in the final Style node of each Layer. This led to no legend image being displayed in the Mapbender Legend element window. This problem was independent of secured vs unsecured, and was also observable pre-pull 616.

Backward scanning will naturally pick the last Style node if it is populated, just like before, so it should be compatible with all previously working cases, and improve legend image picking for previously broken ones.

### DIFF
--- a/src/Mapbender/CoreBundle/Controller/ApplicationController.php
+++ b/src/Mapbender/CoreBundle/Controller/ApplicationController.php
@@ -9,7 +9,7 @@ use Mapbender\CoreBundle\Component\EntityHandler;
 use Mapbender\CoreBundle\Component\SecurityContext;
 use Mapbender\CoreBundle\Entity\Application as ApplicationEntity;
 use Mapbender\CoreBundle\Utils\RequestUtil;
-use Mapbender\WmsBundle\Component\InstanceTunnel;
+use Mapbender\WmsBundle\Component\InstanceTunnelHandler;
 use Mapbender\WmsBundle\Entity\WmsSource;
 use OwsProxy3\CoreBundle\Component\CommonProxy;
 use OwsProxy3\CoreBundle\Component\ProxyQuery;
@@ -387,7 +387,7 @@ class ApplicationController extends Controller
         if (!$requestType) {
             throw new BadRequestHttpException('Missing mandatory parameter `request` in tunnelAction');
         }
-        $instanceTunnel = new InstanceTunnel($instance);
+        $instanceTunnel = new InstanceTunnelHandler($instance);
         $url = $instanceTunnel->getInternalUrl($request);
         if (!$url) {
             throw new NotFoundHttpException('Operation "' . $requestType . '" is not supported by "tunnelAction".');
--- a/src/Mapbender/CoreBundle/Controller/ApplicationController.php
+++ b/src/Mapbender/CoreBundle/Controller/ApplicationController.php
@@ -406,7 +406,7 @@ class ApplicationController extends Controller
                     $glgMode = null;
                     $layerSource = null;
                 } else {
-                    $layerSource = WmsSourceEntityHandler::getLayerSource($source, $layerName);
+                    $layerSource = WmsSourceEntityHandler::getLayerSourceByName($source, $layerName);
                     if (!$layerSource) {
                         $glgMode = null;
                     }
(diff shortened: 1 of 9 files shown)


src/Mapbender/CoreBundle/Controller/ApplicationController.php
```diff
@@ -393,7 +400,28 @@ public function instanceTunnelAction($slug, $instanceId)
                 $url = $source->getGetFeatureInfo()->getHttpGet();
                 break;
             case 'getlegendgraphic':
-                $url = $source->getGetLegendGraphic()->getHttpGet();
+                $glgMode = $request->query->get('_glgmode', null);
```
eSlider: Not the best naming. Hidden request option? :)

src/Mapbender/WmsBundle/Component/WmsInstanceLayerEntityHandler.php
```diff
@@ -331,4 +308,93 @@ private function generateTunnelUrl($url)
...
+        return null;
+    }
+
+    /**
+     * Return the client-facing configuration for a layer's legend
+     *
+     * @param WmsInstanceLayer $entity
+     * @return array
+     */
+    public function getLegendConfig(WmsInstanceLayer $entity)
```
eSlider: Like the split in new methods!

src/Mapbender/WmsBundle/Component/WmsSourceEntityHandler.php
```diff
@@ -154,4 +155,32 @@ public function getInstances()
...
     }
+
+    /**
+     * Find the named WmsLayerSource
+     *
+     * @param WmsSource $source
+     * @param string $layerName
+     * @return WmsLayerSource|null
+     */
+    public static function getLayerSource(WmsSource $source, $layerName)
```
eSlider: getSourceLayerByName - may better?

src/Mapbender/CoreBundle/Controller/ApplicationController.php
```diff
@@ -393,7 +400,28 @@ public function instanceTunnelAction($slug, $instanceId)
...
+                if (!$layerName) {
+                    $glgMode = null;
+                    $layerSource = null;
+                } else {
+                    $layerSource = WmsSourceEntityHandler::getLayerSource($source, $layerName);
+                    if (!$layerSource) {
+                        $glgMode = null;
+                    }
+                }
+                switch ($glgMode) {
```
eSlider: I prefer to create a new method for: `getUrlByMode($mode)`

src/Mapbender/WmsBundle/Component/WmsSourceEntityHandler.php
werrolf: Agreed. Will do.

src/Mapbender/CoreBundle/Controller/ApplicationController.php
werrolf: Agreed. Will do.

src/Mapbender/CoreBundle/Controller/ApplicationController.php
werrolf: "Hidden" because it's definitely not a standard WMS parameter.
glg = short for GetLegendGraphic.

eSlider: ok, I understand.

werrolf:
Tunnel URL processing has been moved out of the controller.

Actually, tunnel URL generation has also been moved.

Commits today add two new Component classes in WmsBundle
* InstanceTunnelHandler evaluates tunnel request url, provides internal URL to satisfy request
* InstanceTunnel (inherits from Handler) that can also generate public URLs to the tunnel action

Hierarchy was split because the full InstanceTunnel requires access to the router, which felt like an excessive dependency for tunnel request handling.

Issue #713, opened 2018-01-10
Plus Button in startpage without login


AxxL:
The "add application" button is visible in the list of applications when you are not logged in. If you click it you have to login. Previous versions haven't shown this button. See attached screenshot.

![mb_306_plus](https://user-images.githubusercontent.com/2235393/34762303-9cd97854-f5e7-11e7-9504-fd2ca086e80b.png)

Please remove this button again if you are not logged in. Show this button only if you are logged in and able to create apps.

werrolf:
Pull ist bereit.

werrolf:
Fix has been approved and merged (into FOM).

werrolf:
Both FOM and Mapbender pulls merged.

AxxL:
OK

PR #714, opened 2018-01-10
Fix false-positive result in SecurityContext::isUserAllowedToCreate(new Application)

werrolf:
This is the mapbender-side fix to issue #713.

We observed a regression around commit 3353075d2c222c2e13f36ff29e3193b455b41edc that enabled a logged-in user with no particular rights (non-root definitely) to successfully create new applications.

I assume the old change was to also support calls to SecurityContext::isUserAllowedToCreate with class name string arguments (as seen [in ExchangeHandler](https://github.com/mapbender/mapbender/blob/e7ee068eb8c494683bb71fc39937f0766df710bf/src/Mapbender/ManagerBundle/Component/ExchangeHandler.php#L96)).

This pull attempts to "normalize" the argument into an ObjectIdentity before forwarding to isGranted. Only the create action check has been touched. For creation, it seems safe to assume a "class" type OI is required because the check has no bearing on any concrete instance. The new method is thus called getClassIdentity.

Behaviour is now as expected: a logged-in user with all assignable privileges revoked cannot create new applications (403 Access denied on /manager/application/new) and is not offered the related interface elements (plus sign above application list, navigation item).
Other grants checks, in particular the one in the ExportHandler, continue to work as before.

### DIFF
--- a/src/Mapbender/CoreBundle/Component/SecurityContext.php
+++ b/src/Mapbender/CoreBundle/Component/SecurityContext.php
@@ -2,6 +2,8 @@
 namespace Mapbender\CoreBundle\Component;
 
 use FOM\UserBundle\Entity\User;
+use Symfony\Component\Security\Acl\Domain\ObjectIdentity;
+use Symfony\Component\Security\Acl\Model\ObjectIdentityInterface;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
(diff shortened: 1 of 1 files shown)


src/Mapbender/CoreBundle/Component/SecurityContext.php
```diff
@@ -220,8 +222,8 @@ public function isUserAnOperator($object)
      */
     public function isUserAllowedToCreate($object)
     {
-        //$oid = new ObjectIdentity('class', get_class($object));
-        return $this->isGranted(self::PERMISSION_CREATE, $object);
+        $identity = $this->getClassIdentity($object);
```
eSlider: The methods is static, so please `static::getClassItedentity($object)` instead.

werrolf: I actually prefer ```$this->``` whenever possible, because calling code doesn't need to change if / when the method changes to non-static (e.g. it needs to access logging or something else via container and can't stay static anymore).

PR #735, opened 2018-02-07
Update element form after submit 

### LINKED ISSUE #734: Update element edit form after submit
When an element edit form (backend) contains invalid data, no error messages are shown after submit because html of the form is not updated.  Also there is no error message if the submit request fails.

sveta-wl:
The html code of an element edit form will be updated after submit using response. All error messages will be shown. Fixes #734

### DIFF
--- a/src/Mapbender/ManagerBundle/Resources/public/js/application-edit.js
+++ b/src/Mapbender/ManagerBundle/Resources/public/js/application-edit.js
@@ -231,10 +231,7 @@ $(function() {
                     label: Mapbender.trans("mb.manager.components.popup.add_element.btn.ok"),
                     cssClass: 'button buttonYes right',
                     callback: function() {
-                        $("#elementForm")
-                                .data('dirty', false)
-                                .submit();
-                        return false;
+                        elementFormSubmit();
                     }
                 },
                 'back': {
@@ -301,9 +298,7 @@ $(function() {
                     label: Mapbender.trans("mb.manager.components.popup.edit_element.btn.ok"),
                     cssClass: 'button right',
                     callback: function() {
-                        $("#elementForm")
-                                .data('dirty', false)
-                                .submit();
+                        elementFormSubmit();
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


src/Mapbender/ManagerBundle/Resources/public/js/application-edit.js
```diff
@@ -324,6 +319,31 @@ $(function() {
...
+            method: 'POST',
+            data: data,
+            error: function (e, statusCode, message) {
+                alert(Mapbender.trans("mb.application.save.failure.general") + ' ' + message);
+            },
+            success: function(data) {
+                if (data.length > 0) {
+                    $(form.parent()).html( data );
+                } else {
+                    $("#elementForm").data('dirty', false);
```
atd-schubert: you have access to the `form` variable in this scope. You don't have to query again.

```js
form.data('dirty', false);
```

src/Mapbender/ManagerBundle/Resources/public/js/application-edit.js
```diff
@@ -324,6 +319,31 @@ $(function() {
         return false;
     });
 
+    function elementFormSubmit() {
+        var form = $("#elementForm"),
```
atd-schubert: This is just a personal favor, but maybe prefix elements in their jQuery variant with a `$`.

src/Mapbender/ManagerBundle/Resources/public/js/application-edit.js
```diff
@@ -324,6 +319,31 @@ $(function() {
...
+            data = form.serialize(),
+            url = form.attr('action'),
+            self = this;
+
+        $.ajax({
+            url: url,
+            method: 'POST',
+            data: data,
+            error: function (e, statusCode, message) {
+                alert(Mapbender.trans("mb.application.save.failure.general") + ' ' + message);
```
atd-schubert: prevent using `alert()` in JavaScript! Don't we have a better way to display errors in Mapbender?

I mean there was something like: `Mapbender.error()`...

src/Mapbender/ManagerBundle/Resources/public/js/application-edit.js
```diff
@@ -324,6 +319,31 @@ $(function() {
...
+            },
+            success: function(data) {
+                if (data.length > 0) {
+                    $(form.parent()).html( data );
```
atd-schubert: I think `from.parent().html( data );` will also work...

src/Mapbender/ManagerBundle/Resources/public/js/application-edit.js
```diff
@@ -324,6 +319,31 @@ $(function() {
...
+                } else {
+                    $("#elementForm").data('dirty', false);
+                    self.close();
+                    window.location.reload();
```
atd-schubert: Why do you reload the page here? Is it really necessary? When it is necessary, is there an advantage in closing the form before?

werrolf: Because the usual server response to a successful POST (or PUT; we don't support form methods, apparently!) is to redirect to the modified resource view with a 303.

We kind of "emulate" the traditional behavior here.

Issue #736, opened 2018-02-07
Colors of Zoombar and Toolbar are different


AxxL:
If you compare the colors of the zoombar and toolbar in the standard fullscreen template, you'll see that there are small differences:

![bild_ticket_uniform_color](https://user-images.githubusercontent.com/2235393/35924631-d7b57944-0c23-11e8-968d-3474cafbd0e1.png)

The zoom-in button and zoom-to-world-extent button is slightly darker than the history_prev/history_next button.

Next is the toolbar. You'll see also, that these are a bit darker than the rest. The next screenshot shows the difference. Please note, that I have adjusted the color of the zoomin/zoomout buttons by hand in Firebug (in difference to the previous screenshot)

![bild_ticket_uniform_color_mit_angepasster_zoomleiste](https://user-images.githubusercontent.com/2235393/35924734-25e83d9a-0c24-11e8-9141-a49dbbbc94ae.png)

I think it's nicer if they are all lighter and that we should use the lighter alternative as a template.

I expect them to have all the same color.

**Note** It should be done for 3.0.6 AND 3.0.5.

werrolf:
Fix available in release/3.0.5 and release/3.0.6

PR #737, opened 2018-02-07
Changed Opacity for zoombar and toolbar to get a unique button color

AxxL:
Changed the opacity of the zoombar and toolbar to make it unique.

https://github.com/mapbender/mapbender/issues/736


werrolf:
The 305 version of the branch has been merged, this one is no longer needed.

PR #741, opened 2018-02-09
Clean the code of the elementFormSubmit function

sveta-wl:
Fixed comments from the review https://github.com/mapbender/mapbender/pull/735#pullrequestreview-94707821. Relates to the issue #734


sveta-wl:
@atd-schubert without `!important` the message is shown under the dark background hover of the modal.

PR #748, opened 2018-02-15
Feature/introspections for sources <=> applications <=> sourcinstances

werrolf:
New CLI command mapbender:inspect:source:usage

Place a `new MapbenderInspectionBundle(),` into your AppKernel if you don't get it automagically.

Shows relations between applications and sources and source instances (`--by-app`, default) or the other way around (`--by-source`).

Specially highlights disabled source instances and unpublished applications.

Completely unused sources are listed separately.

I've built the tabular display first, but the collection components are completely separate from the CLI command, and should be useable as is for e.g. HTML rendering somewhere in a backend extension.
The generated aggregates hold all involved entities in full.

### DIFF
--- a/src/Mapbender/IntrospectionBundle/Command/SourcesCommand.php
+++ b/src/Mapbender/IntrospectionBundle/Command/SourcesCommand.php
@@ -9,9 +9,9 @@ use Mapbender\IntrospectionBundle\Component\Aggregator\Relation\ApplicationToSou
 use Mapbender\IntrospectionBundle\Component\Aggregator\Relation\SourceToApplications;
 use Mapbender\IntrospectionBundle\Component\Aggregator\Source;
 use Mapbender\IntrospectionBundle\Component\Collector;
-use Mapbender\IntrospectionBundle\Entity\Utils\Command\DataGroup;
+use Mapbender\IntrospectionBundle\Entity\Utils\Command\DataTreeNode;
 use Mapbender\IntrospectionBundle\Entity\Utils\Command\DataItem;
-use Mapbender\IntrospectionBundle\Entity\Utils\Command\DataRootGroup;
+use Mapbender\IntrospectionBundle\Entity\Utils\Command\DataItemList;
 use Mapbender\IntrospectionBundle\Entity\Utils\Command\JsonFormatting;
 use Mapbender\IntrospectionBundle\Entity\Utils\Command\YamlFormatting;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
@@ -77,7 +77,7 @@ class SourcesCommand extends ContainerAwareCommand
             'Sources',
             'Instances',
         );
-        $tree = new DataRootGroup();
+        $tree = new DataItemList();
         foreach ($aggregate->getRelations(true) as $appInfo) {
             $tree->addItem($this->collectAppRelation($appInfo));
         }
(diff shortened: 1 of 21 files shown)


werrolf:
Actually, I'd like some added YAML output myself, so the result can be stored at different points in a  project's development and compared with a diff more easily.

The current table output is fancy, but if one column changes width, you'd get an "all lines have changed" situation in a diff.

src/Mapbender/IntrospectionBundle/Component/WorkingSet.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function getSources()
+    {
+        return $this->sources;
+    }
+
+    /**
+     * @param Application[]|null $applications
+     */
+    public function setApplications(array $applications = null)
```
sveta-wl: Do we actuallly need null here? May be just assume that Applications is always an array and skip the further checking?

src/Mapbender/IntrospectionBundle/Component/WorkingSet.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function addSource(Source $source)
+    {
+        $this->sources[] = $source;
+    }
+
+    /**
+     * @param Source[]|null $sources
+     */
+    public function setSources(array $sources = null)
```
sveta-wl: The same question as by setApplication

src/Mapbender/IntrospectionBundle/Command/SourcesCommand.php
```diff
@@ -0,0 +1,247 @@
...
+
+        $this->buckets = array();
+        $noteStyle = new OutputFormatterStyle('white', 'blue');
+        $output->getFormatter()->setStyle('note', $noteStyle);
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $collector = new Collector($this->getContainer());
+        if ($this->bucketBy == 'application') {
```
sveta-wl: May be it's possible to divide this function for smaller ones to increase reading comfortability?

src/Mapbender/IntrospectionBundle/Component/WorkingSet.php
werrolf: Actually, the entire WorkingSet thing doesn't seem as useful as hoped, because it's only checked to initialize a single entity list. Everything else is ignored.

It was easier to follow the defined Doctrine relations.

A better mechanism than this WorkingSet construct would be pluggable filters, that could e.g. strip disabled (or enabled) source instances, strip unpublished applications, or filter based on text matching the title.

That's why I kind of "forgot" to maintain the WorkingSet halfway through :p

src/Mapbender/IntrospectionBundle/Command/SourcesCommand.php
werrolf: Yes.

I also want to break the table formatting stuff out of that command, because it's used verbatim in other commands and should be in a central place.

src/Mapbender/IntrospectionBundle/Component/WorkingSet.php
werrolf: I think the original idea was to treat "null" as "do whatever you do by default" vs empty array = "you should really process the empty set of applications".
It's true that the implementation doesn't reflect that and it doesn't look worthwhile atm.

PR #757, opened 2018-02-16
Dimension handler: fix lost config in frontend

werrolf:
There was some "Update the code" commit that introduced a loop over the dimension configuration in layersets, and emptied it, by reference.

This pull fixes that. Dimension configration survives app initialization.


werrolf:
Reasons lost in history. The temp[0] thing has been introduced in commit 68b0c3ab6b8d8c0b00b323bff3aa58e2419a2a3e "change frontend of DimensionsHandler"

Issue #756, opened 2018-02-16
Form validators for backend


sveta-wl:
The element forms in the backend must have validation. 
It needs to provide relevant validators and constraints for checking form data.

werrolf:
Tested:
- [x] Enforcement of required constraints
- [x] Enforcement of HTML validity constraints
- [x] Saving Element after logging out in second tab redirects to login page, then resumes the save after successful login

Close?

PR #758, opened 2018-02-16
Enhancement/form validators

sveta-wl:
This pull request relates to the issue https://github.com/mapbender/mapbender/issues/756 and contains:
-  declaration form types as services to be able to inject there any other necessary/desired services
- initialization of form type by taking it from the container using its service id
- declaration of constraints as services
- provided constraints and validators to validate html and twig content by HTMLElement setup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 * **v.3.0.6.4**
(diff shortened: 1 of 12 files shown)


AxxL:
@sveta-wl please note it in the changelogs. thanks.

PR #765, opened 2018-02-21
Fix/wms cleanups loading

werrolf:
**NOTE:** This trojan-horses pull #764.

This pull moves the Wms loading machinery out of the WmsBundle RepositoryController actions into an "Importer" component that is demonstrably reusable.
It also introduces
* a new (and trivial) WmsOrigin Entity that bundles url and the optional user + password for basic auth protected services
* a new (and similarly trivial) Importer Response that carries the parsing results along with an (optional) ignored validation exception, so it can still be displayed as a warning if the user chose to ignore structural errors in the source's GetCapabilities

[Used like this:](https://github.com/mapbender/mapbender/commit/a73d3a5e28085c85d6375378e13036fede232d1a#diff-1e3841a9d0469ca05262f37ad58ac2beR176)
```php
$importer = new Importer($this->container);  // @todo: symfony service :p
$origin = new WmsOrigin($url, $userName, $password);
$importerResponse = $importer->evaluateServer($origin, false);
/** @var \Exception|null $validationError */
$validationError = $importerResponse->getValidationError();
/** @var \Mapbender\WmsBundle\Entity\WmsSource $wmsSource */
$wmsSource = $importerResponse->getWmsSourceEntity();
```

This pull also resolves some issues concerning storage of temporary schema downloads in the XmlValidator. These used to go into the web folder, which may not even be writable depending on deployment details. Worse, if web wasn't writable, an infinite loop was triggered.
The solution was to distinguish "shipping" schemas (packed in with mapbender, considered static and an optimization) from "ad-hoc" schemas that must be downloaded because the document that is currently being validated references them.

The "ad-hoc" schemas now go into `sys_get_temp_dir().'/mapbender/xmlschemas'`. Better than web. They are deleted when processing completes, either successfully or throwing.
The "shipped" schemas remain in web/xmlschemas where they have been. I have confirmed that they are still picked up and used.

### CLI / testing
As a proof of concept, there is a new mapbender:wms:validate:url CLI command that accepts an URL. It either bubbles any occuring exceptions, or trivially displays the detected layers when all went well.

I have tested both "DTD" and "non-DTD" strict parsing and evaluation, and both work.

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Exception/InvalidUrlException.php
+++ b/src/Mapbender/CoreBundle/Component/Exception/InvalidUrlException.php
@@ -1,0 +1,13 @@
+<?php
+
+
+namespace Mapbender\CoreBundle\Component\Exception;
+
+
+class InvalidUrlException extends \InvalidArgumentException
+{
+    public function __construct($url)
+    {
+        parent::__construct("Invalid url " . var_export($url, true));
+    }
+}
--- a/src/Mapbender/CoreBundle/Component/XmlValidator.php
+++ b/src/Mapbender/CoreBundle/Component/XmlValidator.php
@@ -195,12 +195,51 @@ EOF
                 /** @var \DOMElement $import */
                 $ns_ = $import->getAttribute("namespace");
                 $sl_ = $import->getAttribute("schemaLocation");
-                $this->addSchemaLocationReq($schemaLocations, $ns_, $sl_);
+                $schemaUrl = $this->resolveRelativeUrl($ns, $sl_);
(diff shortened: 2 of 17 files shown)


werrolf:
[Most recent commit](https://github.com/mapbender/mapbender/pull/765/commits/dffb846632261423294efab976ae0dd113e00159) improves reuse of shipping schemas named after legacy conventions (flat directory, lots of underscores).

In testing this shaves off 500ms from a 1.5s total validation + loading time of a pretty trivial 1.3.0 WMS (3 layers).

The only upstream change to the schemas since we downloaded them in 2014 was reordering the `targetNamespace="http://www.opengis.net/wms" ` attribute to go first in the `<schema>` node.

PR #770, opened 2018-02-22
Fix #698 load WMS covers redlining

### LINKED ISSUE #698: load WMS / WMC covers redlining
A WMS added with WMS or WMC loader covers the redlining layer.
Redlinging-Layer should be pushed to the top after loading additional WMS.

see:
https://release.mapbender.org/application/wmcredlining

wms to add:
http://ows.terrestris.de/osm/service?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetCapabilities

DeEgge:
The redlining element now listens on event 'mbmapsourceadded'. If this is triggered, the redlining layer is raised by the number of current layers and the layerZIndex is reset

See ticket: https://github.com/mapbender/mapbender/issues/698

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
@@ -35,17 +35,14 @@
             if(this.options.auto_activate || this.options.display_type === 'element'){
                 this.activate();
             }
+
+            this.setupMapEventListeners();
+
             this._trigger('ready');
             this._ready();
-
-            //put rl layer on top of layer stack if a source is added, i.e. by wms loader
-            var self = this;
-            $(document).on('mbmapsourceadded', function(event, params) {
-                if (self.layer) {
-                    self.map.raiseLayer(self.layer, self.map.getNumLayers());
-                    self.map.resetLayersZIndex();
-                }
-            });
+        },
+        setupMapEventListeners: function() {
+            $(document).on('mbmapsourceadded', this._moveLayerToLayerStackTop.bind(this));
         },
         defaultAction: function(callback){
             this.activate(callback);
@@ -313,6 +310,16 @@
                     $('.geometry-table tr[data-id="'+this.selectedFeature.id+'"] .geometry-name', this.element).text(label);
                     this.layer.redraw();
                 }
+            }
+        },
+        /**
+         * Move redlining layer on top of layer stack if a source is added, i.e. by wms loader
+         * @private
+         */
+        _moveLayerToLayerStackTop: function(event, params) {
+            if (this.layer) {
+                this.map.raiseLayer(this.layer, this.map.getNumLayers());
+                this.map.resetLayersZIndex();
             }
         },
         /**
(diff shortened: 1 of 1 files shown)


src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
```diff
@@ -37,6 +37,15 @@
             }
             this._trigger('ready');
             this._ready();
+
+            //put rl layer on top of layer stack if a source is added, i.e. by wms loader
+            var self = this;
```
LazerTiberius: Would be better to use .bind(this); 
```
function(event, params) {
    if (this.layer) {
       this.map.raiseLayer(this.layer, this.map.getNumLayers());
       this.map.resetLayersZIndex();
    }
}.bind(this)
```
Instead of referencing this

src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
```diff
@@ -37,6 +37,15 @@
             }
             this._trigger('ready');
             this._ready();
+
+            //put rl layer on top of layer stack if a source is added, i.e. by wms loader
```
sveta-wl: I would prefer to setup such event listeners in a separate function. If it needs a comment for the following lines - please put these lines in a separate function also. This way the reading of the code and the widget inheritance will be much easier.

werrolf:
I'm still not entirely sure about this, but ok for now.

In the testing application, this of course produces the desired behavior.
But I'm wondering what happens if you combine multiple elements with highlighting or other geometry layers
* print allows you to pull up a rectangle
* ruler / area ruler allow you to pull up marker layers
* the map itself lets you draw a rectangle on top to zoom in
* Redlining, Digitizer are all about drawing geometries on top of the map

Allowing an element to start its own fight over the z index, and wanting to go to the top of the stack seems like a workaround for a lack of central logic and management of layers.

Strategically, the map should manage these things and distinguish between opaque grid layers ("base sources"), transparent grid layers, static vector and then highlight layer, and a) keep them stacked in that order, b) conceptually guarantee that "the active element" (in an application with accordions or tabs) gets the top feature visibility.

Then we wouldn't need to manage this on a per-element basis anymore.

DeEgge:
Maybe this is the wrong place for a discussion, but here's some food for thought:
Why don't we handle vector layers via the layertree? It's the most common thing in GI systems to have every geometry in a separate layer and to be able to rearrange the order of these layers.
Temporary geometries like rulers, printing rectangles and the zoom rectangle should always be on top of the layer stack.
The order of Digitizer and Redlining could be manageable via the layertree.
Highlighting layers could either reveal them, even if they are underneath another layer (to give the user a hint on where they are located), or do nothing at all...

DeEgge:
I've investigated this a bit further.
The 'hold-shift-and-zoom'-box will always be on top of everything, no matter what.
If you use the line or area ruler and draw a vector layer AND keep the element active, it will also be covered by a loaded WMS. However, turning it off and on again, will result in the usual behavior (measure vector on top of WMS).
With this fix, redlining will not cover the measure tools (I'm guessing OpenLayer.Handler.Path will also always be on top of everything else).

werrolf:
General consensus seemed to be to go through with this until the need for a better solution becomes more than theoretical.

PR #774, opened 2018-02-26
update example WMS to https and version 1.3.0

astroidex:
* change example WMS to https and change Version to 1.3.0


werrolf:
[Already done](https://github.com/mapbender/mapbender/commit/d379a8d16d0eab15d91f8235745618052ae28a99) :p

PR #775, opened 2018-02-26
Fix/add axis order performance

werrolf:
[The mapaxisorder.js file](https://github.com/mapbender/mapbender/blob/4095eb58d6b81ec8513e2d7b31e0b1a562ae3578/src/Mapbender/CoreBundle/Resources/public/mapbender.element.map.mapaxisorder.js) contains a list of 1885 EPSG codes for which the WMS 1.3 BBOX axis order reverseral applies.

The Chrome JavaScript profiler tells me the original version took 10ms to evaluate.

What I did in the first commit was to run a multiline regex search and replace

from ```true\}\s*\},\s*\{\s*\"``` to ```true\},\n\t\t\"```

Then I had one object.
Next I eliminated the loop.

This brings the Profiler timing down to 1.6ms.


werrolf:
[The next commit](5483f2fe63fac9a80ce65bcf3050b82d2ffb45f8) is potentially unsafe. Basically I preinitialize one object, and assign it to every code.
```js
    var yxTrue = {yx: true};

    var codes = {
        "EPSG:2036": yxTrue,
        "EPSG:2044": yxTrue,
        "EPSG:2045": yxTrue,
        "EPSG:2065": yxTrue,
<...>
```
This halves runtime again to 800µs.

This has an "interesting" side effect though. All code objects are the same now, by reference. You can set _an_ _object_ to different value(s) and the replacement works as expected.
```
> OpenLayers.Projection.defaults["EPSG:31283"] = {yx: false}
{yx: false}
> OpenLayers.Projection.defaults["EPSG:31278"]
{yx: true}
```
If you alter the yx property on one of them, all of them are changed (because they actually all are the same).
```
> OpenLayers.Projection.defaults["EPSG:31277"]
{yx: true}
OpenLayers.Projection.defaults["EPSG:31278"].yx = false
false
OpenLayers.Projection.defaults["EPSG:31277"]
{yx: false}
```

Is this safe? Is there a plausible reason for alterations to happen at runtime? If they happen, do they access the property directly, or do they replace the whole object?

LazerTiberius:
Should be safe. There is no plausible reason to alter any object after the initial setup. This is just an extension because OL2 only defines these values for a handful projections and without these WMS 1.3 services would be requested with a wrong axis order.

PR #790, opened 2018-03-12
Fix/wms cleanups entity behaviour

werrolf:
This is a backwards-compatible change dissolving dependencies on EntityHandler while interacting with sources, source instances, their layers, and related objects.

The public APIs of all EntityHandler child classes has been maintained including all quirks. Old-style instantiation and usage is fully supported, many repository-internal usages have been updated though.

### Motivation
The EntityHandler hierarchy is an embodiment of three anti-patterns that make maintaining and extending the behaviors incredibly complex
* explicitly runtime-instantiated service
* non-reusable service prebound to single task
* intransparent string name concatenation of concrete class

### Instantiation, IDE introspection
EntityHandler instantiation has been simplified and clarified, replacing instances of `EntityHandler::createHandler` with a plain old `new` whenever the resulting class is deterministic, and a move to a more concrete e.g. `SourceInstanceEntityHandler::createHandler` wherever some polymorphism is possible. This greatly improves IDE type knowledge, making the using code easier to understand and maintain.

Superfluous instantiations of EntityHandler have also been removed wherever possible. E.g. [internal cases of EntityHandler classes creating, basically, themselves again have been reduced to simple recursive calls](https://github.com/mapbender/mapbender/pull/790/files#diff-df3c2648c88e6cc0566723726ad7a60fL173)

## Recursive relational object deletion without EntityHandler interaction
[EntityHandler-driven recursive remove has been cut back](https://github.com/mapbender/mapbender/pull/790/commits/d13a4c73729dabe4ca5b49b6898cebb6f39fa699). This is already ensured by cascade=remove declarations in the entities. This change has been verified against services with populated `contact` and `keyword` relations.
* Tables have been counted
* complex WMS source has been added, instance has been attached to application
* Source has been deleted.
* I observed that the instance vanished, all attached layers and all attached keywords and contacts vanished. The table counts matched exactly the state before adding the source.

## Freeing up WmsInstance::configuration
The [last important change](https://github.com/mapbender/mapbender/commit/a44a91aa62100215ab0bc6bc6f8d7604766b5a32) is stopping frontend config being smeared into WmsInstance::configuration, previously requiring, again, creation of a WmsInstanceEntityHandler and a call to generateConfiguration whenever something relating to frontend config had changed.

This was effectively a caching layer built into a core entity, which was a horrenduous anti-pattern. The [payload we had been caching](https://github.com/mapbender/mapbender/blob/fbb9e44f0a3344b3d07e6d8525f9ec29bab49bce/src/Mapbender/WmsBundle/Component/WmsInstanceConfigurationOptions.php#L328) was the result of
* calling ~20 getters on the WmsInstance
* serializing Bounding Boxes, Dimensions and Vendorspecifics to arrays
* converting ratio and opacity to float

This payload is ~constant time, it does not grow with the layer count, only slightly with Dimensions. It's not worth breaking entity semantics to cache it, and if we still need to cache it, that needs to happen in the presentation layer, perhaps in the scope of caching the entire application config (TBD).

Generated configuration is the same as before. For demonstration, I baked a few monster WMS services into an application and diffed the config 1) with the code right before this branch and 2) at the (current) end of this branch with the added extra of **never** returning the persisted configuration from before from WmsInstance::getConfiguration. I'm using json_pp with `canonical` option to ensure JSON keys are sorted and don't appear in random order, so the configs are comparable.
```
## release/3.0.5, parent commit of this branch
ron@flake:~/proj/mb305/application/mapbender$ git checkout -q ef7e083e39d82324cd04281beda06188e168c6e2
ron@flake:~/proj/mb305/application/mapbender$ rm -rf `find ../app/cache/{dev,prod} -mindepth 1 -maxdepth 1 -not -name 'sessions'`
ron@flake:~/proj/mb305/application/mapbender$ curl -s 'http://305.mapbender.local/app_dev.php/application/teh-grids/config' | json_pp -json_opt=pretty,ascii,canonical > teh-grids.config.ef7e083e39d82324cd04281beda06188e168c6e2.json

## last config-significant commit on this branch
ron@flake:~/proj/mb305/application/mapbender$ git checkout -q fbb9e44f0a3344b3d07e6d8525f9ec29bab49bce

## manually disable WmsInstance::getConfiguration (not committed) ..., then
ron@flake:~/proj/mb305/application/mapbender$ git diff src/
diff --git a/src/Mapbender/WmsBundle/Entity/WmsInstance.php b/src/Mapbender/WmsBundle/Entity/WmsInstance.php
index daf58bd..26b0d14 100644
--- a/src/Mapbender/WmsBundle/Entity/WmsInstance.php
+++ b/src/Mapbender/WmsBundle/Entity/WmsInstance.php
@@ -201,6 +201,7 @@ class WmsInstance extends SourceInstance
      */
     public function getConfiguration()
     {
+        return array();
         return $this->configuration;
     }

ron@flake:~/proj/mb305/application/mapbender$ rm -rf `find ../app/cache/{dev,prod} -mindepth 1 -maxdepth 1 -not -name 'sessions'`
ron@flake:~/proj/mb305/application/mapbender$ curl -s 'http://305.mapbender.local/app_dev.php/application/teh-grids/config' | json_pp -json_opt=pretty,ascii,canonical > teh-grids.config.fbb9e44f0a3344b3d07e6d8525f9ec29bab49bce.json

## now finally compare the configs
ron@flake:~/proj/mb305/application/mapbender$ diff -uw teh-grids.config.a44a91aa62100215ab0bc6bc6f8d7604766b5a32.json teh-grids.config.fbb9e44f0a3344b3d07e6d8525f9ec29bab49bce.json
ron@flake:~/proj/mb305/application/mapbender$
```
The diff is empty => the configs delivered to the client are 100% unchanged.

The pretty-printed json configuration is ~780kB (8 active WMS instances). The local WmsInstance configurations made up 2.5% of the data volume and 12% of the total generation time. The practical impact of no longer caching these portions is smaller, because the instance is now more compact and loads faster.

### Strategy
The ultimate goal (beyond the scope of this pull) is to no longer depend on the EntityManager hierarchy at all. Recursive persistence, removal, attaching related objects can and should be performed entirely by Doctrine EntityManager. What the EM cannot do should move into compartmentalized services (proper services). Namely:
* reformatting entity data into a configuration array fit for the frontend (presentation layer)
* merging a new entity into an old entity, e.g. refresh WmsSource in backend updates related WmsInstances (Repository layer)

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Application.php
+++ b/src/Mapbender/CoreBundle/Component/Application.php
@@ -334,7 +334,7 @@ class Application
 
             foreach ($layerSet->layerObjects as $layer) {
                 /** @var SourceInstanceEntityHandler|WmsInstanceEntityHandler $instHandler */
-                $instHandler = EntityHandler::createHandler($this->container, $layer);
+                $instHandler = SourceInstanceEntityHandler::createHandler($this->container, $layer);
                 $conf        = $instHandler->getConfiguration($this->container->get('signer'));
 
                 if (!$conf) {
--- a/src/Mapbender/CoreBundle/Component/Application.php
+++ b/src/Mapbender/CoreBundle/Component/Application.php
@@ -7,6 +7,7 @@ use Mapbender\CoreBundle\Component\Element as ElementComponent;
 use Mapbender\CoreBundle\Entity\Application as Entity;
 use Mapbender\CoreBundle\Entity\Element as ElementEntity;
 use Mapbender\CoreBundle\Entity\Layerset;
+use Mapbender\WmsBundle\Component\WmsInstanceEntityHandler;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
@@ -332,6 +333,7 @@ class Application
             $layerSets     = array();
 
             foreach ($layerSet->layerObjects as $layer) {
+                /** @var SourceInstanceEntityHandler|WmsInstanceEntityHandler $instHandler */
                 $instHandler = EntityHandler::createHandler($this->container, $layer);
                 $conf        = $instHandler->getConfiguration($this->container->get('signer'));
 
--- a/src/Mapbender/CoreBundle/Component/Base/ConfigurationBase.php
+++ b/src/Mapbender/CoreBundle/Component/Base/ConfigurationBase.php
@@ -1,0 +1,105 @@
+<?php
+
+
+namespace Mapbender\CoreBundle\Component\Base;
+
+
+/**
+ * Base class for simple "plain old data" configuration carriers that support
+ * * conversion TO array
+ * * instantiation and prepopulation FROM array
+ *
+ * Going back and forth should be idempotent. fromArray has a (by default enabled) strict mode that
+ * enforces that only known value keys can appear in the input array.
+ */
+abstract class ConfigurationBase
+{
+    protected static $classDefaults = array();
+
+    /**
+     * Convert the instance to an array representation.
+     *
+     * @return mixed[]
+     */
+    abstract public function toArray();
+
+    /**
+     * Factory method. Instantiates class and prepopulates it with values from given $options.
+     *
+     * @param mixed[] $options
+     * @param boolean $strict to throw an Exception if unrecognized options were passed
+     * @return static
+     * @throws \InvalidArgumentException if $options is not an array
+     * @throws \RuntimeException if unrecognized keys are found in $options
+     */
+    public static function fromArray($options, $strict = true)
+    {
+        if (!is_array($options)) {
+            if ($strict) {
+                throw new \InvalidArgumentException('Options must be an array, is: ' . print_r($options, true));
+            } else {
+                $options = array();
+            }
+        }
+        $instance = new static();
+        $instance->populateAttributes($options, $strict);
+        return $instance;
+    }
+
+    /**
+     * Perform mass attribute population from array via magic attribute access.
+     *
+     * @param mixed[] $options
+     * @param boolean $strict
+     * @throws \RuntimeException if unsupported value keys are found and $strict == true
+     */
+    protected function populateAttributes($options, $strict)
+    {
+        if ($strict) {
+            $validateAgainst = $this->validSet();
+            $badKeys = array_keys(array_diff_key($options, $validateAgainst));
+            if ($badKeys) {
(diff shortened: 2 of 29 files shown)


src/Mapbender/CoreBundle/Component/Application.php
```diff
@@ -332,7 +333,8 @@ public function getConfiguration()
             $layerSets     = array();
 
             foreach ($layerSet->layerObjects as $layer) {
-                $instHandler = EntityHandler::createHandler($this->container, $layer);
+                /** @var SourceInstanceEntityHandler|WmsInstanceEntityHandler $instHandler */
+                $instHandler = SourceInstanceEntityHandler::createHandler($this->container, $layer);
```
sveta-wl: I would prefer to have here an injected service to increase decoupling, then call of a static method of a class, if possible

src/Mapbender/CoreBundle/Component/Base/ConfigurationBase.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * @return array
+     * Policy; no "get" prefix for non-serializible getter functions in objects that are persisted
+     * or exported or otherwise magically scanned for methods with "get" / "is" / "has" etc
+     */
+    public static function validSet()
+    {
+        $cls = get_called_class();
```
sveta-wl: Maybe a better name for $cls, like $calledClass or so?

src/Mapbender/CoreBundle/Component/Application.php
werrolf: This branch is pretty old! Config service has already been implemented. This whole section is gone after the merge commit.

src/Mapbender/CoreBundle/Component/Base/ConfigurationBase.php
werrolf: This whole class structure turned out to be transient, i.e. it helped me understand (and unit-test) what was going on. It's a candidate for deprecation + removal now, along with the WmsInstanceConfiguration and WmsInstanceConfigurationOptions.

werrolf:
Work in this area is not complete, but there are cross-dependencies between changes in this branch and other developments (config service etc), so to continue reasonably we need a starting point containing both.

We still haven't achieved the goal of a simple $em->persist, $em->flush interaction, but that's the goal of course.

PR #791, opened 2018-03-12
Fix/wms cleanups freeform instance config

werrolf:
This is a continuation of [pull #790](https://github.com/mapbender/mapbender/pull/790). [Here's a clean diff without the base changes](https://github.com/mapbender/mapbender/compare/fbb9e44f0a3344b3d07e6d8525f9ec29bab49bce...37b5f4ac60db495c0c73943e4f0c3132c89f61f9).

**WARNING WARNING DANGER DANGER**
Saving any WmsInstance on this branch will strip the "configuration" column. Which is not a problem if you continue working on this version, but you cannot go back to an older version. Back up your tables.

This warning would become obsolete after a merge of 790.
**WARNING WARNING DANGER DANGER**

Ok.

This commit repurposes WmsInstance::configuration to become a wild-west storage attribute for unstructured extensions, just like Element::configuration. This allows us to deliver extended instance configuration in cases where we would previously need to add a column, and require running a doctrine:schema:update. Such as
* layer order reversal
* UtfGrid specifics (tile size override, layer whitelist)

Unlike 790, this is a breaking change as it modifies database values in a way that precludes a downgrade back to any previous versions. This will warrant a version bump and a special mention in release notes, recommendation to make database backups etc.

I'm putting this here more as an RFC. If I could add a blocking review myself, I would. This will however become the base for certain client-specific fast-track developments.

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Application.php
+++ b/src/Mapbender/CoreBundle/Component/Application.php
@@ -334,7 +334,7 @@ class Application
 
             foreach ($layerSet->layerObjects as $layer) {
                 /** @var SourceInstanceEntityHandler|WmsInstanceEntityHandler $instHandler */
-                $instHandler = EntityHandler::createHandler($this->container, $layer);
+                $instHandler = SourceInstanceEntityHandler::createHandler($this->container, $layer);
                 $conf        = $instHandler->getConfiguration($this->container->get('signer'));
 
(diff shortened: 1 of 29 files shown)


werrolf:
After some internal discussions this week, I'm closing this. We're going to prefer added columns via doctrine:schema:update.

PR #794, opened 2018-03-15
Add config hook to expose DimensionsHandler as a technical preview

werrolf:
This is a better, cleaner version of [pull 610](https://github.com/mapbender/mapbender/pull/610/files#diff-005c46c7584cee49cf60354ae9ad5ac0L47), exploring a new "technical preview" delivery for not-quite-ready features that are already in the main code base.

DimensionsHandler has been sleeping in the code base for a while, but it's not ready yet for public consumption. Some projects do use it though, even on older 3.0.5 versions, so we can't just remove it or break it out into its own repository without risking upgrade continuity.

This pull does not enable anything new by default. DimensionsHandler is still kept "hidden". Meaning:
* doesn't show up in the list of addable elements when you edit your application layout
* doesn't show up in LayerTree "menu" choices

This simply changes the method of exposure. Before:
* Projects would have to hack their MapbenderWmsBundle::getElements
* Projects would have to hack their Element\Type\LayerTreeAdminType

This may actually have led to forks.

After:
* Projects can at their own risk enable DimensionsHandler it with a single parameters.yml entry
```yml
parameters:
    <...>
    mapbender.preview.element.dimensionshandler: true
```

### DIFF
--- a/src/Mapbender/CoreBundle/Element/Type/LayerTreeMenuType.php
+++ b/src/Mapbender/CoreBundle/Element/Type/LayerTreeMenuType.php
@@ -1,0 +1,61 @@
+<?php
+
+
+namespace Mapbender\CoreBundle\Element\Type;
+
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+
+class LayerTreeMenuType extends AbstractType
+{
+    protected $exposedChoices = array();
+
+    /**
+     * @param boolean $showDimensions see services.xml, controlled via parameter `mapbender.preview.element.dimensionshandler`
+     */
+    public function __construct($showDimensions)
+    {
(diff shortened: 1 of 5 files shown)


werrolf:
Pinging @CharlyUnicorn @astroidex for general awareness.

astroidex:
@werrolf - sounds good to me. Thanks for pinging.

CharlyUnicorn:
@werrolf this is a nice way to handle the dimensions element - but in my opinion additionally to the code base you need to deliver a documentation of the feature. 

first version of documentation: 
* https://github.com/mapbender/mapbender-documentation/blob/feature/dimensions-handler/en/functions/misc/dimensions_handler.rst
* https://github.com/mapbender/mapbender-documentation/blob/feature/dimensions-handler/de/functions/misc/dimensions_handler.rst

werrolf:
Documentation is a separate repository, and such any updates will have to be done and tracked separately :\

Issue #785, opened 2018-03-09
[BaseSourceSwitcher] with group - group is opened on start of the application and modified click behaviour


astroidex:
* occurs actual Mapbender from 2018-03-08
* BaseSourceSwitcher group dropdown is opened on start of application (this is a behaviour that is not wanted) -> looks strange with many entries (overlaps map)
* On click on a different menu the first one does not close

### Demo Application the check the behaviour

- https://release.mapbender.org/application/Geoportal__BaseSourceSwitcher_785

### How can you reproduce the error?

- open the application -> wrong behaviour:  Basesourceswitcher is opened already
- check what happens when you click on the menu items
- click on OSM -> OSM mit Opacity -> wrong behaviour -> menu is still open

### Check application to see how it should work
- https://luftbilder.geoportal.ruhr/

![mb_bs1](https://user-images.githubusercontent.com/996298/37211797-658ac382-23ad-11e8-937c-24b4f629e4db.png)
![mb_bs2](https://user-images.githubusercontent.com/996298/37211803-6aac2a54-23ad-11e8-8340-5ab5a9f3191d.png)

AxxL:
Das Beispiel nutzt Gruppen im im BaseSourceSwitcher. 

- Die 3 ersten Instancesets gehören zur einer Group "nrw"
- Die nächsten 2  gehören zu einer Group "osm".
- kein Hintergrund hat keine BaseSource ausgewählt. Und keine Gruppe.
- WebAtlAS Graustufen ist einzeln und gehört zu keiner Gruppe.

# Beobachtete Fehler:

- Bei diesen Menüs, die dann durch die Groups entstehen wird das erste (hier nrw bestehend aus DOP2 et al) automatisch geöffnet. Wenn man das zweite Menü öffnet (hier OSM), dann überdeckt das das erste.

Ich gehe davon aus, dass der Fehler nur darin besteht, dass das erste Menü (NRW) automatisch geöffnet wird.

CharlyUnicorn:
When I upgraded Mapbender in January 2018, I noticed the same misbehavior:
Using a group in the BaseSourceSwitcher always opens the group at application startup.
Only by clicking on an entry, the group closes.
Testet on an intern Instance at https://bev-dev.wheregroup.com/mapbender/app_dev.php

werrolf:
Can't reproduce in 3.0.5, seems like purely a 3.0.6 issue (which will make it easier to fix).

DeEgge:
[...]

PR #798, opened 2018-03-19
Fix basesourceswitcher automatically opening on startup

DeEgge:
This will prevent the basesource switcher from opening on application startup.
Fixes https://github.com/mapbender/mapbender/issues/785

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.basesourceswitcher.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.basesourceswitcher.js
@@ -27,6 +27,10 @@
                 } else {
                     bsswtch.addClass('hidden');
                 }
+
+                if (!e.hasOwnProperty('originalEvent')) {
+                    bsswtch.addClass('hidden');
+                }
             });
 
             this._hideSources();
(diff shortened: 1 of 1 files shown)


src/Mapbender/CoreBundle/Resources/public/mapbender.element.basesourceswitcher.js
```diff
@@ -17,33 +17,28 @@
...
 
-            this.element.find('.basesourcesetswitch:first').click();
-
             $(document).on('mbmapcontextaddstart', $.proxy(this._onContextAddStart, this));
             $(document).on('mbmapsourceloadstart', $.proxy(this._onSourceLoadStart, this));
             $(document).on('mbmapsourceloadend', $.proxy(this._removeSourceFromLoad, this));
             $(document).on('mbmapsourceloaderror', $.proxy(this._removeSourceFromLoad, this));
         },
 
+        _showHideMenu: function(e) {
```
atd-schubert: Do you mean `_toggleMenu`?

But this fix is much better then the one before (in my opinion...)!

werrolf:
Do we need a changelog? Was this a regression since 3.0.6.3 or was the issue introduced at some point after the release?

PR #811, opened 2018-04-16
Fix/target element type robustness

werrolf:
This is development QoL.

Avoids errors when editing applications that may contain Elements not present in the code (on the current branch etc).

When editing any Element that can control a "target" (Button, FeatureInfo etc), the backend generates a list of potential targets by scanning all Elements (including inactive ones) present in the Application.

Previously, any dangling class references would lead to an exception. I.e. editing any controlling-type Element in the backend "Layouts" section would fail building the form with an HTTP 500 status.

This pull introduces a change mirroring already present existance checks for the concrete Element class  found elsewhere:
* Layout section itself displays the Element with an empty title
* Application frontend rendering and config generation skips processing the Element altogether

This pull only touches one class (the TargetElementType form type), which is identical between 3.0.5 and 3.0.6. Because the class is barely getting touched, I took the opportunity to do some c&p folding and fix the formatting.

Dummy Application component instantiation was redundant and has been removed.

The significant logic change is entirely in [the final commit](https://github.com/mapbender/mapbender/pull/811/commits/a682cc2b53eb687c4c70447f87a9cfb89822330d).

### DIFF
--- a/src/Mapbender/CoreBundle/Element/Type/TargetElementType.php
+++ b/src/Mapbender/CoreBundle/Element/Type/TargetElementType.php
@@ -84,9 +84,11 @@ class TargetElementType extends AbstractType
         /** @var EntityRepository $repository */
         $repository = $this->getContainer()->get('doctrine')->getRepository($options['class']);
         $qb = $repository->createQueryBuilder($builderName);
+        $applicationFilter = $qb->expr()->eq($builderName . '.application', $options['application']->getId());
         $filter = $qb->expr()->andX();
+        $filter->add($applicationFilter);
+
         if (!empty($options['element_class'])) {
-            $filter->add($qb->expr()->eq($builderName . '.application', $options['application']->getId()));
             if (is_integer(strpos($options['element_class'], "%"))) {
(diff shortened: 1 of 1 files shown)


src/Mapbender/CoreBundle/Element/Type/TargetElementType.php
```diff
@@ -99,9 +98,12 @@ public function getChoicesQueryBuilder(Options $options)
...
             $elm_ids = array();
             foreach ($options['application']->getElements() as $element_entity) {
-                $class = $element_entity->getClass();
-                if ($class::$ext_api) {
-                    $element_entity->getId();
+                /** @var Element $element_entity */
+                $elementComponentClass = $element_entity->getClass();
+                if (class_exists($elementComponentClass)) {
+                    if ($elementComponentClass::$ext_api) {
+                        $element_entity->getId();
```
LazerTiberius: Would be a  great opportunity to remove snake_case and move to a consistent lower camel case.

werrolf: Agreed, will do.

PR #812, opened 2018-04-17
Feature/application config cache

werrolf:
This plugs a performance regression introduced in the context of plugging security issues related to "instance tunneling" getLegendGraphic requests.

We introduce machinery to cache arbitrary application-specific data. The cache is self-managing / self-invalidating. Based on Application `updated` and time of the last container compilation, values are stale with no outside interaction (i.e. no deletion of cache files). Get is fully atomic. Put is nearly atomic, save for a potential mkdir vs external deletion race, but any errors this might produce are caught and processing is stopped.

Config caching is generally less aggressive than the existing caching systems for js and css assets, so the amount of manual cache clears should not increase.

Config caching can be disabled via parameter `cachable.mapbbender.application.config`.

On my X230, config response latency improves anywhere between 40% and 850%, depending on application complexity. Configs are identical (md5sum and all that). Configs still update as expected when editing the application in the backend. Yaml applications update immediately in app_dev, and require a (Symfony) cache clear in "prod" (same as before).

### monster application (144kB config)
```
ron@flake:~$ wrk -c 2 -t 2 -d 20s 'http://305.mapbender.local/app.php/application/teh-grids/config'
Running 20s test @ http://305.mapbender.local/app.php/application/teh-grids/config
  2 threads and 2 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    94.61ms    5.69ms 175.83ms   99.29%
    Req/Sec    10.57      2.42    20.00     93.47%
  422 requests in 20.02s, 59.64MB read
Requests/sec:     21.08
Transfer/sec:      2.98MB
```
### small application (10kB config)
```
ron@flake:~$ wrk -c 2 -t 2 -d 20s 'http://305.mapbender.local/app.php/application/mapbender_mobile/config'
Running 20s test @ http://305.mapbender.local/app.php/application/mapbender_mobile/config
  2 threads and 2 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    13.67ms    1.12ms  32.01ms   95.80%
    Req/Sec    73.19      6.35    80.00     56.75%
  2928 requests in 20.01s, 27.38MB read
Requests/sec:    146.30
Transfer/sec:      1.37MB
```

### monster application (144kB config)
```
ron@flake:~$ wrk -c 2 -t 2 -d 20s 'http://305.mapbender.local/app.php/application/teh-grids/config'
Running 20s test @ http://305.mapbender.local/app.php/application/teh-grids/config
  2 threads and 2 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.96ms    1.18ms  26.11ms   91.96%
    Req/Sec   100.85      9.62   111.00     79.25%
  4022 requests in 20.03s, 564.37MB read
Requests/sec:    200.79
Transfer/sec:     28.18MB
```

### small application (10kB config)
```
ron@flake:~$ wrk -c 2 -t 2 -d 20s 'http://305.mapbender.local/app.php/application/mapbender_mobile/config'
Running 20s test @ http://305.mapbender.local/app.php/application/mapbender_mobile/config
  2 threads and 2 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.19ms    1.24ms  40.81ms   94.18%
    Req/Sec   109.39      9.57   121.00     78.50%
  4360 requests in 20.01s, 40.77MB read
Requests/sec:    217.88
Transfer/sec:      2.04MB
```

## Related
Basing cache invalidation also on container compilation time goes beyond existing caching mechanisms for js / css assets, which only look at the `updated` time on the Application entity, but don't consider system configuration changes.

Those systems are also pretty much hard-wired to Assetic and Twig embedding, making them difficult to expand upon.

Strategically, caching systems should unify on top of the simplest implementation.

The machinery introduced here is basic and bare bones and makes few assumptions about intended usage. The cache deals only with raw strings and offers prepackaged retrieval as Response object only as an extra convenience.

Caching of StringAsset types etc could easily be layered on top of this.

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
+++ b/src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
@@ -17,8 +17,19 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  */
 class File
 {
+    // constants related to internal header construction
+    const H_LENGTHOF_SIGNATURE_LENGTH = 6;
+    const H_LENGTHOF_VALUE_LENGTH = 6;
+
     /** @var string */
     protected $rootPath;
+
+    /**
+     * Lazy-initialized sprintf pattern used to generate the header plus value we place / expect in the file.
+     * @var string|null
+     */
+    protected $storagePattern;
+
 
     public function __construct(ContainerInterface $container)
     {
@@ -44,10 +55,16 @@ class File
         // 1) achieve complete atomicity of put / get
         // 2) be independent of various mtime resolutions on various filesystems when using time stamps
         // 3) support arbitrary (any serializable) signatures instead of just timestamps
+        // We also encode the length of the signature and the data itself in a fixed-length fashion, to ease reading
 
-        // Encode the length of the signature and the data itself in a fixed-length fashion, to ease reading
-        $prefix = sprintf("%6d:%6d:%s:", strlen($signature), strlen($value), $signature);
-        $valueInternal = $prefix . $value;
+        if (!$this->storagePattern) {
+            $this->storagePattern = '%' . static::H_LENGTHOF_SIGNATURE_LENGTH . 'd'
+                                  . ':'
+                                  . '%' . static::H_LENGTHOF_VALUE_LENGTH . 'd'
+                                  // two more placeholders for signature + value
+                                  . ':%s:%s';
+        }
+        $valueInternal = sprintf($this->storagePattern, strlen($signature), strlen($value), $signature, $value);
         @file_put_contents($fullPath, $valueInternal, LOCK_EX);
     }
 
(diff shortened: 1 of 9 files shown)


src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
```diff
@@ -0,0 +1,103 @@
...
+class File
+{
+    /** @var string */
+    protected $rootPath;
+
+    public function __construct(ContainerInterface $container)
+    {
+        /** @var BaseKernel $kernel */
+        $kernel = $container->get('kernel');
+        $rootPath = realpath($kernel->getCacheDir());
```
sveta-wl: It's also possible to inject the kernel.cache_dir path directly from the service configuration (services.xml) as an argument:
```
<service id="mapbender.presenter.application.cache.backend" class="%mapbender.presenter.application.cache.backend.class%">
<argument>%kernel.cache_dir%</argument> 
</service>
```
and get rid of $container and $kernel variables. It could be more readable this way, what do you think?

src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
```diff
@@ -0,0 +1,103 @@
...
+     * @param string[] $keyPath a multi-component key, in this backend used as a subdirectory hierarchy
+     * @param string $value will be stored exactly as given
+     * @param mixed $mark to determine reusability of the cache entry later in get
+     */
+    public function put($keyPath, $value, $mark)
+    {
+        // prepend root path
+        array_unshift($keyPath, $this->rootPath);
+        $fullPath = implode('/', $keyPath);
+        @mkdir(dirname($fullPath));
```
sveta-wl: Are you sure it's a good idea to use @? I always have a bad feeling regarding things like this...

src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+        $header = substr($valueInternal, 0, 14);
+        $markLength = intval(substr($header, 0, 6));
+        $markSerialized = serialize($mark);
+        $markPrevious = substr($valueInternal, 14, $markLength);
+        if ($markPrevious === false || $markPrevious !== $markSerialized) {
+            throw new CacheMiss();
+        }
+        $valueLength = intval(substr($header, 7, 6));
+        $valueStart = 14 + $markLength + 1;
```
sveta-wl: May be use constants, which explain what all these numbers mean? It's quite confusing to me

src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
```diff
@@ -0,0 +1,102 @@
...
+            throw new CacheMiss();
+        }
+        $header = substr($valueInternal, 0, 14);
+        $signatureLength = intval(substr($header, 0, 6));
+        $cachedSignature = substr($valueInternal, 14, $signatureLength);
+        if ($cachedSignature === false || $cachedSignature !== $signature) {
+            throw new CacheMiss();
+        }
+        $valueLength = intval(substr($header, 7, 6));
+        $valueStart = 14 + $signatureLength + 1;
```
sveta-wl: Maybe use constants, which explain what all these numbers mean? It's quite confusing to me

src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
werrolf: Sounds good, I'll do it.

werrolf: [Done](https://github.com/mapbender/mapbender/pull/812/commits/5a0865e376ed52ed0d57a8b7434eba3bbad8b710)

src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
werrolf: It's just a cache. None of its contents are required for the application to function. Everything that goes into the cache can be generated again.
As such, I'm ignoring interaction errors in put silently so the application can keep running.

Likewise, errors from file_put_contents are ignored as well.

The worst possible end result of errors in either mkdir or file_put_contents is a cache miss when the system looks for reusable data. Cache misses are also not error conditions.

src/Mapbender/CoreBundle/Component/Cache/Backend/File.php
werrolf: @LazerTiberius also took issue with the magic values, and [I later added some class contents and lots of comments to explain what's happening](https://github.com/mapbender/mapbender/pull/812/commits/47566afb599dcace3c682c239748c6120d5d6ca5).

PR #821, opened 2018-04-27
Fix coordinate font size NaN

### LINKED ISSUE #817: [PrintClient] print error with PHP 7.1 as fontsize for extent_ur_y usw. runs on an error
Steps to reproduce:
* make a print pdf print on Windows Apache 2.4 with PHP 7.1 
 

Observed behaviour:

* pdf ist not generated
* exit with notice Notice: A non well formed numeric value encountered

```
in vendor\setasign\fpdf\fpdf.php at line 516  -+     $this->FontFamily = $family; 
    $this->FontStyle = $style; 
    $this->FontSizePt = $size; 
    $this->FontSize = $size/$this->k; 
    $this->CurrentFont = &$this->fonts[$fontkey]; 
    if($this->page>0) 
        $this->_out(sprintf('BT /F%d %.2F Tf ET',$this->CurrentFont['i'],$this->FontSizePt));
```

`at PrintService ->addCoordinates () 
in mapbender\src\Mapbender\PrintBundle\Component\PrintService.php at line 485  -+         if (isset($this->conf['fields']['extent_ur_x']) && isset($this->conf['fields']['extent_ur_y']) 
                && isset($this->conf['fields']['extent_ll_x']) && isset($this->conf['fields']['extent_ll_y'])) 
        { 
            $this->addCoordinates(); 
        } 
`
* see actual master: https://github.com/mapbender/mapbender/blob/master/src/Mapbender/PrintBundle/Component/PrintService.php#L644

If you define a fixed font size the error does not occur anymore an pdf is generated
* f.e.
`        $pdf->SetFont('Arial', '', 6 );`

* reproduced the error with David on Windows Apache 2.4 with PHP 7.1

DeEgge:
When printing a template with coordinates being printed, the font size was not parsed as an integer, but as a string. This fixes #817

### DIFF
--- a/src/Mapbender/PrintBundle/Component/PrintService.php
+++ b/src/Mapbender/PrintBundle/Component/PrintService.php
@@ -641,25 +641,25 @@ class PrintService extends ImageExportService
         }
 
         // upper right Y
-        $pdf->SetFont('Arial', '', $this->conf['fields']['extent_ur_y']['fontsize']);
+        $pdf->SetFont('Arial', '', intval($this->conf['fields']['extent_ur_y']['fontsize']));
         $pdf->Text($this->conf['fields']['extent_ur_y']['x'] + $corrFactor,
                     $this->conf['fields']['extent_ur_y']['y'] + 3,
                     round($this->data['extent_feature'][2]['y'], $precision));
 
         // upper right X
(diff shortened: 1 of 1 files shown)


LazerTiberius:
Till Php7.0 Php has cast something like "15pt"*4 happily to (int)60. From 7.1 Php now raises a notice.

Issue #814, opened 2018-04-25
When creating the new application the given screenshot image is not used


AxxL:
Steps to reproduce:

- Create new application
- Choose Fullscreen Template
- Choose a screenshot image
- Save

Observed behaviour:
- The screenshot image that I have chosen is not displayed. I have to edit the application and upload it again, then it works.

Expected behaviour:
- The screenshot image is automatically uploaded and chosen when I create the new application in the first step.

PR #824, opened 2018-05-04
Fix/814 sreenshot image bug on create application

inacht:
#814: When creating the new application the given screenshot image is not used.


inacht:
Sorry, the commit messages are confusing. For solving the issue "!$form->submit($request)->isValid()"
was relevant (instead of "!$form->submit($parameters)->isValid()" )  .

inacht:
@sveta-wl  ok, i have done the changes as wished. Also i did change the way to get the request and submitting the form data (werrolf did mention that it was done the deprecated way, therefore i did change it).

PR #846, opened 2018-06-01
Move WMS layer order default application from entity => entity handle…

werrolf:
Remove default for null columns from Entity getLayerOrder.

A/B testing on Postgres revealed that for reasons not entirely understood the entity does not update in the database if the only changed column is layerOrder as long as the getter provides a default. Removing the default (i.e. getLayerOrder returns plain attribute, whatever it is) restores reliable database updates.

Add back the (configurable) default into the frontend config emission machinery.


werrolf:
I don't think it's the same cause. The issue here was that the value that controls reversal of all layers would not save to the database.
The other issue is that runtime modifications (drag&drop, WmsLoader) in the layertree are not handled reliably.

Issue #830, opened 2018-05-17
[featureInfo] FEATURE_COUNT=10 is set for every FeatureInfo-Request


astroidex:
**Problem**
* the FeatureInfo-Request give back results a a click pixel position. It can happen that there are more than one results a that position
* Mapbender reduces the number of results to get from the WMS to 10 in the FeatureInfo-Request (... &REQUEST=GetFeatureInfo&FEATURE_COUNT=10 ...)
* This is not good and can assume a wrong impresion to the user if there are more results ther in reality

**Steps to reproduce**
* make a FeatureInfo request in an application
*  F12 and get the Request-Url and look for FEATURE_COUNT
* you will see that FEATURE_COUNT IS set to 10 fixed

**Example**
> https://release.mapbender.org/owsproxy/?url=https://wms.wheregroup.com/cgi-bin/mapbender_user.xml?_signature=53%3AHCOOjVz8_jkxOCNplHhyjzueKV8&LAYERS=Mapbender_Names&QUERY_LAYERS=Mapbender_Names&STYLES=&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetFeatureInfo&BBOX=360194.96972%2C5619969.989546%2C371955.03028%2C5624030.010454&FEATURE_COUNT=10&HEIGHT=580&WIDTH=1680&FORMAT=image%2Fpng&INFO_FORMAT=text%2Fhtml&CRS=EPSG%3A25832&I=697&J=267

**Change request**
* please increase the value to at least 100 or 200. It will set a higher maximum for results to give back

astroidex:
* assigned to @werrolf 
* I reviewd the PR. works fine. Thanks

PR #849, opened 2018-06-07
Raise maxFeatures to 1000

LazerTiberius:
https://github.com/mapbender/mapbender/issues/830


astroidex:
* reviewed. Works fine. Would be great to have the fix in 3.0.7.x

PR #852, opened 2018-06-12
[WIP] dynamic layer reordering vs reversed layer order

werrolf:
Status: dynamic reordering of both static and dynamically added sources and layers works as intended. "Themed" layertree was tested as well.
Layer order self-reset on scroll / dynamic layer de-/reactivation has been resolved.

We had some issues with dynamic layer reordering in combination with "Reversed (QGIS style)" layer order in some applications.

The previous code performed a multitude of incremental "Move this layer before this second layer and / or after that third layer" updates to the internal "geosource" data structures, in hopes that the internal structure would always remain in sync with the (visible) DOM order. This was often not the case. Particularly, when many layers were moved in bulk, the before / after layers could point to already moved ones, breaking the process.

The changes on this branch make setting the desired layer order per source a one-shot bulk operation, and remove the incremental single-layer "move" operations.

The layertree was adapted to use the bulk layer order setter on the geosource, so the result will remain consistent with the visible DOM order of the layers.
The "sourceMoved" event is still fired but without the old structured "changed" payload, which would have been excrutiatingly hard to emulate. The only known consumer of this event (inside our own codebase) is the legend widget, which [does not access this information at all](https://github.com/mapbender/mapbender/blob/v3.0.7.1/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js#L105).

The layer ordering ended up as global state (keyed on source id), which is fugly. I resorted to this because JOII would not let me access parent-declared object properties in a child instance...

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
@@ -184,8 +184,8 @@
             $('li.leave[data-type="simple"]', $sourceContainer).each(function() {
                 var $t = $(this);
                 var layerId = $t.attr('data-id');
-                if (layerId !== undefined) {
-                    layerIdOrder.push(layerId.toString());
+                if (typeof layerId !== "undefined") {
+                    layerIdOrder.push("" + layerId);
                 }
             });
             this.model.setSourceLayerOrder(sourceId, layerIdOrder.reverse());
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
@@ -217,10 +217,10 @@
                         if (type === 'theme' || type === 'root') {
                             self._updateSourceOrder();
                             // self._sortTheme($elm);
-                        } else if ($elm.attr('data-type') === 'simple') {
+                        } else if (type === 'simple' || type === 'group') {
                             self._updateSource($elm.closest('.serviceContainer'));
                         } else {
-                            console.warn("Warning: unhandled element in layertree sorting", $elm);
+                            console.warn("Warning: unhandled element in layertree sorting", type, $elm);
                         }
                     }
                 });
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
@@ -188,7 +188,7 @@
                     layerIdOrder.push(layerId.toString());
                 }
             });
-            this.model.setSourceLayerOrder(sourceId, layerIdOrder);
+            this.model.setSourceLayerOrder(sourceId, layerIdOrder.reverse());
         },
         /**
          * Applies the new (going by DOM) ordering between sources.
@@ -540,7 +540,6 @@
         _removeChild: function(changed) {
             var self = this;
             if (changed && changed.sourceIdx && changed.childRemoved) {
-                var source = this.model.getSource(changed.sourceIdx);
                 $('ul.layers:first li[data-id="' + changed.childRemoved.layer.options.id + '"]', self.element).
                     remove();
             }
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
@@ -170,100 +170,37 @@
         },
         _unSortable: function() {
         },
-        _sortItem: function($toMoveItem, $beforeItem, $afterItem) {
-            var before = null;
-            var after = null;
-            var toMove = null;
-            if ($beforeItem) {
-                if ($beforeItem.hasClass('themeContainer')) {
-                    var $beforeSEl = $beforeItem.find('ul.layers:first li[data-type="root"]:last');
-                    var bid = $beforeSEl.attr('data-sourceid');
-                    before = {
-                        sourceIdx: {
-                            id: bid
-                        },
-                        layerIdx: {
-                            id: $beforeSEl.attr("data-id")
-                        }
-                    };
-                } else {
-                    var bid = $beforeItem.attr('data-sourceid');
-                    bid = bid ? bid : $beforeItem.parents('li[data-sourceid]:first').attr('data-sourceid');
-                    before = {
-                        sourceIdx: {
-                            id: bid
-                        },
-                        layerIdx: {
-                            id: $beforeItem.attr("data-id")
-                        }
-                    };
-                }
-            }
-            if ($afterItem) {
-                if ($afterItem.hasClass('themeContainer')) {
-                    var $afterSEl = $afterItem.find('ul.layers:first li[data-type="root"]:first');
-                    var bid = $afterSEl.attr('data-sourceid');
-                    after = {
-                        sourceIdx: {
-                            id: bid
-                        },
-                        layerIdx: {
-                            id: $afterSEl.attr("data-id")
-                        }
-                    };
-                } else {
-                    var bid = $afterItem.attr('data-sourceid');
-                    bid = bid ? bid : $afterItem.parents('li[data-sourceid]:first').attr('data-sourceid');
-                    after = {
-                        sourceIdx: {
-                            id: bid
-                        },
-                        layerIdx: {
-                            id: $afterItem.attr("data-id")
-                        }
-                    };
-                }
-            }
-            var id = $toMoveItem.attr('data-sourceid');
-            id = id ? id : $toMoveItem.parents('li[data-sourceid]:first').attr('data-sourceid');
-            toMove = {
-                sourceIdx: {
-                    id: id
-                }
-            };
-            if ($toMoveItem.attr("data-type") !== this.consts.root) {
-                toMove['layerIdx'] = {
-                    id: $toMoveItem.attr("data-id")
-                };
-            }
-            this.model.changeSource({
-                change: {
-                    move: {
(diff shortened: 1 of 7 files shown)


src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -170,100 +170,37 @@
...
+         */
+        _updateSource($sourceContainer) {
+            // this will capture the "configurationish" layer ids (e.g. "1_0_4_1") from
+            // all layers in the source container in DOM order
+            var sourceId = $sourceContainer.attr('data-sourceid');
+            var layerIdOrder = [];
+            $('li.leave[data-type="simple"]', $sourceContainer).each(function() {
+                var $t = $(this);
+                var layerId = $t.attr('data-id');
+                if (layerId !== undefined) {
```
LazerTiberius: Please stick to one way of undefined check. Further in the code typeof varX  !== "undefined" is used which is the safer way to do.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -170,100 +170,37 @@
...
+                    layerIdOrder.push(layerId.toString());
```
LazerTiberius: Please stick to one way of int to string cast. Further in the code ```""+int``` is used which is the faster way to do.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -276,10 +213,14 @@
                     cursor: "move",
                     update: function(event, ui) {
                         var $elm = $(ui.item);
-                        if ($elm.hasClass('themeContainer')) {
-                            self._sortTheme($elm);
+                        var type = $elm.attr('data-type');
+                        if (type === 'theme' || type === 'root') {
```
LazerTiberius: switch case would be easíer to extend in the future

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -276,10 +213,14 @@
...
+                            self._updateSourceOrder();
+                            // self._sortTheme($elm);
+                        } else if (type === 'simple' || type === 'group') {
+                            self._updateSource($elm.closest('.serviceContainer'));
```
LazerTiberius: Please use bind instead of creating a reference to the outer scope this

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -599,7 +540,6 @@
         _removeChild: function(changed) {
             var self = this;
```
LazerTiberius: Use of self is redundant

src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
```diff
@@ -85,7 +85,7 @@
             widget.onMapLayerChanges();
 
             $(document)
-                .bind('mbmapsourceadded mbmapsourcechanged mbmapsourcechanged mbmapsourcemoved', $.proxy(widget.onMapLayerChanges, widget))
+                .bind('mbmapsourceadded mbmapsourcechanged mbmapsourcemoved', $.proxy(widget.onMapLayerChanges, widget))
```
LazerTiberius: Please use bind instead of $.proxy

src/Mapbender/CoreBundle/Resources/public/mapbender.geosource.js
```diff
@@ -243,12 +246,12 @@ Mapbender.Geo.SourceHandler = Class({
                     if (layer.children[i].options.id.toString() === offsetLayer.options.id.toString()) {
                         options.found = true;
                         if (options.cut_with) {
```
LazerTiberius: Variable names and object keys should be always lover camel case

src/Mapbender/CoreBundle/Resources/public/mapbender.geosource.js
```diff
@@ -243,12 +246,12 @@ Mapbender.Geo.SourceHandler = Class({
...
-                            var lays = layer.children.splice(i, layer.children.length - i);
+                            var lays = layer.children.slice().splice(i, layer.children.length - i);
```
LazerTiberius: Maybe use a better name for the variable lays. What does lays mean?

src/Mapbender/CoreBundle/Resources/public/mapbender.geosource.js
```diff
@@ -718,6 +728,13 @@ Mapbender.Geo.SourceHandler = Class({
             return source.configuration.options.bbox;
         }
         return null;
+    },
+    'public function setLayerOrder': function(source, layerIdOrder) {
+        var self = this;
+        Mapbender.Geo.layerOrderMap["" + source.id] = $.map(layerIdOrder, function(layerId) {
+            var layerObj = self.findLayer(source, {id: layerId});
```
LazerTiberius: Please use bind instead of creating a reference to the outer scope this

src/Mapbender/CoreBundle/Resources/public/mapbender.model.js
```diff
@@ -1084,118 +1043,67 @@ Mapbender.Model = {
...
+        });
+        // Collect currently set positions and z indexes for given sources.
+        // position := array index in this.sourceTree
+        // z index := mapquery layer position = openlayers map layer index - 1
+        // The collected values will be reused / redistributed to the affected
+        // sources.
+        var oldPositions = [];
+        var zIndexes = [];
+        var sourceIdToSource = {};
+        $.each(sourceObjs, function(i, sourceObj) {
```
LazerTiberius: For better performance, it recommends using underscore/lodash or better native functions if no DOM is involved

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
werrolf: Can't forward `this` via `bind` here, because the `this` in the inner loop scope is the matched `ul.layers` element. If I clobber that with the widget reference, the code would no longer function.

I'm removing the "that" though.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
werrolf: Code not touched, same as before, avoiding mixing in too much refactoring with a behaviour fix.

src/Mapbender/CoreBundle/Resources/public/mapbender.geosource.js
werrolf: Code not touched, same as before, avoiding mixing in too much refactoring with a behaviour fix.

werrolf: ... buuuuut renamed to `ìncludeOffset` now. That's the name of the value in the object passed in to getLayersList. Kind of makes sense to keep it constant through the call chain, and it's more intuitive this way what it's doing.

src/Mapbender/CoreBundle/Resources/public/mapbender.geosource.js
werrolf: Whole section cleaned up in [
af77e22](https://github.com/mapbender/mapbender/pull/852/commits/af77e2272abda2a742eaa7bef1839b3c04f25c71).

src/Mapbender/CoreBundle/Resources/public/mapbender.geosource.js
werrolf: [Done](https://github.com/mapbender/mapbender/pull/852/commits/19cb907bfc8007a3dde800a46d91abfbf0b7b2e9)

src/Mapbender/CoreBundle/Resources/public/mapbender.model.js
werrolf: [Done](https://github.com/mapbender/mapbender/pull/852/commits/bd838ead0b4c8de5dca79a8136b32fe53148f31e)

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
werrolf: [Done](https://github.com/mapbender/mapbender/pull/852/commits/c992508c9fca4dd591d0652255f0d5e4b334b7fc)

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
werrolf: [Done?](https://github.com/mapbender/mapbender/pull/852/commits/c992508c9fca4dd591d0652255f0d5e4b334b7fc)

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
werrolf: [Wish granted](https://github.com/mapbender/mapbender/pull/852/commits/58247f30d7e8021ceb566c21d0ef5d6ac5ad92df)

werrolf:
@LazerTiberius I'm going to super-cow merge now because we need the behavioural fix for the imminent new version.  
Most of your concerns have been adressed, but not all.

I would prefer to postpone further and wider-reaching QoL / style / efficiency enhancements, which are clearly, clearly necessary, to a separate pull at a later time.

PR #934, opened 2018-06-19
Modify redlining according to airbnb styles

DeEgge:


src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
```diff
@@ -1,11 +1,11 @@
...
-    $.widget("mapbender.mbRedlining", {
+    $.widget('mapbender.mbRedlining', {
         options: {
             target: null,
-            display_type: 'dialog',
-            auto_activate: false,
-            deactivate_on_close: true,
+            displayType: 'dialog',
+            autoActivate: false,
+            deactivateOnClose: true,
```
werrolf: Those values come from the Element configuration, stored in the DB. You can't just rename them here.  
Even if you rename them equivalently in the Element form, all existing, running applications would effectively lose their configuration for this element.

src/Mapbender/CoreBundle/Resources/public/mapbender.element.redlining.js
```diff
@@ -29,10 +30,11 @@
         },
         _setup: function(){
             this.map = $('#' + this.options.target).data('mapbenderMbMap').map.olMap;
+            this.model = new Mapbender.Model();
```
werrolf: Why create a new instance? You are not creating your entire own layer stack, you cooperate with the main map.  
Use `this.map.model` to access the same model instance the main map uses.

PR #935, opened 2018-06-19
Add methods for getting controls and interactions by classname

DeEgge:


werrolf:
The methods don't work, they always return false :)  
I don't think we'll use this, sorry.

PR #952, opened 2018-06-25
basic imageExport functionality

inacht:
basic imageExport functionality done and some new functions in the model.

### DIFF
--- a/src/Mapbender/PrintBundle/Resources/public/mapbender.element.imageExport.js
+++ b/src/Mapbender/PrintBundle/Resources/public/mapbender.element.imageExport.js
@@ -1,11 +1,11 @@
 (function($){
-
-    $.widget("mapbender.mbImageExport", {
+    'use strict';
+    $.widget('mapbender.mbImageExport', {
         options: {},
         map: null,
         popupIsOpen: true,
-        _create: function(){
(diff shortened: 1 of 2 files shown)


src/Mapbender/PrintBundle/Resources/public/mapbender.element.imageExport.js
```diff
@@ -88,95 +94,50 @@
...
-                    extentheight: mapExtent.getHeight(),
-                    vectorLayers: vectorLayers
-                };
-                var form = $('<form method="POST" action="' + url + '"  />');
-                $('<input></input>').attr('type', 'hidden').attr('name', 'data').val(JSON.stringify(data)).appendTo(form);
-                form.appendTo($('body'));
-                form.submit();
-                form.remove();
-            }
+            var form = $('<form method="POST" action="' + url + '"/>');
```
werrolf: I know you just transformed this from the old code, but we should really earmark this for a change.  
I suppose we could simply ajax the entire data as is, without using an ephemeral form at all?  
If not, at least attributes should be set with safe methods. Sth like
```
var $form = $('<form>');
$fom.attr('action', url);
$form.attr('method', 'POST');
```

PR #955, opened 2018-06-26
Fix ol4 FeatureInfo

philvanbwg:


werrolf:
Works fine.  
Bit of a hack with the mqLayer. We'll clean this up later.

PR #956, opened 2018-06-26
Add return source object from collectFeatureInfoObjects()

philvanbwg:
Neben `label` und `id` wird doch auch noch `source.configuration.options.tunnel` benötigt für Passwort-geschützte Dienste: 

https://github.com/mapbender/mapbender/blob/ol4/src/Mapbender/CoreBundle/Resources/public/mapbender.element.featureInfo.js#L190

Der Ajax-Request sieht dann anders aus


werrolf:
I'm not sure this option is needed client-side. This might be a legacy construct.  
The tunnel endpoint URL is generated server-side. The client doesn't need to know about this at all.  

I will test with a password-protected service to tet.

PR #957, opened 2018-06-26
Ol4 image export with geometries

inacht:
image export now considering vector layer geometries


werrolf:
Can't get this working even with no geometries at all.
Export without any geometry throws a HTTP 500 server-side.   
```
Warning: imagecreatetruecolor() expects parameter 1 to be integer, string given
```

[Site of error](https://github.com/mapbender/mapbender/blob/04b0951616890635aada5ba27088437bfd3ca2d3/src/Mapbender/PrintBundle/Component/ImageExportService.php#L122) (ImageExportService L122)

Please enable PHP warnings for all development.  

Speculation: width + height are posted to server as strings, need to be integers.

PR #973, opened 2018-06-29
handling empty style objects when generating/getting vector layer print style options

inacht:
Handling cases when a  style Object of a vector layer has empty fill, stroke and text objects.


werrolf:
Export / print [already have](https://github.com/mapbender/mapbender/blob/bdeb37a42918efe5b1dad0155b472466f6edebe6/src/Mapbender/PrintBundle/Component/LayerRendererGeoJson.php#L110) all the default styles they need (which definitely do not include hover-specific styles) and follows [the defined defaults](http://dev.openlayers.org/releases/OpenLayers-2.13.1/docs/files/OpenLayers/Feature/Vector-js.html#OpenLayers.Feature.Vector.Constants) more closely than this pull.

Issue #977, opened 2018-07-04
Redlining: "Stealth Mode" Bug


rschwammborn:
It seems that the Redlining feature has several bugs.

Firstly, drawing geometries is possible, but almost all types disappear from the map after the drawing process is done.
The geometries become invisible and are still there, as it is possible to let them reappear via pressing the "edit" button. Zooming and deleting is also possible ( - but they are only visible in edit mode).
Exception: "Text (annotations)" works properly and does not disappear.

Secondly, switching the SRS disrupts the geometries' location. They change based on the chosen EPSG-code. Relations between them stay the same (distance, shape, etc.)

DeEgge:
> Firstly, drawing geometries is possible, but almost all types disappear from the map after the drawing process is done.
The geometries become invisible and are still there, as it is possible to let them reappear via pressing the "edit" button. Zooming and deleting is also possible ( - but they are only visible in edit mode).
Exception: "Text (annotations)" works properly and does not disappear.

671447c should fix this issue

> Secondly, switching the SRS disrupts the geometries' location. They change based on the chosen EPSG-code. Relations between them stay the same (distance, shape, etc.)

@RobinSchwammborn Could you please open a new issue for this? Thanks!

DeEgge:
> Secondly, switching the SRS disrupts the geometries' location. They change based on the chosen EPSG-code. Relations between them stay the same (distance, shape, etc.)

I've just noticed this is the exact same behaviour as with ol2, because the is no transformation happening.

AxxL:
Redlinig doesn't support SRS change

AxxL:
Reopen because of unmerged #980

werrolf:
Redlining OpenLayers 2 base was a completely broken, barely functional mess at the start of the initial OL4 push and should never even have started the porting process.  
The OL2 base is in much, much better shape now. The previous OL4 porting attempt results have lost their value.

PR #980, opened 2018-07-05
Fix invisible features on drawend

DeEgge:
This fixes the first part of #977.

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.model.ol4.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.model.ol4.js
@@ -434,7 +434,9 @@ Mapbender.Model.prototype.createVectorLayer = function(options, owner) {
     var uuid = Mapbender.UUID();
     this.vectorLayer[owner] = this.vectorLayer[owner] || {};
     options.map = this.map;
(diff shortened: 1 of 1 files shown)


src/Mapbender/CoreBundle/Resources/public/mapbender.model.ol4.js
```diff
@@ -434,7 +434,9 @@ Mapbender.Model.prototype.createVectorLayer = function(options, owner) {
     var uuid = Mapbender.UUID();
     this.vectorLayer[owner] = this.vectorLayer[owner] || {};
     options.map = this.map;
-    options.style = options.style ? this.createVectorLayerStyle(options.style) : this.createVectorLayerStyle();
+    options.style = !(options.style instanceof ol.style.Style) ? this.createVectorLayerStyle(options.style) :
```
LazerTiberius: Please save logic expression in a separated variable for better readability e.G. var isOl4Styl = options.style instanceof ol.style.Style

werrolf:
Redlining OpenLayers 2 base was a completely broken, barely functional mess at the start of the initial OL4 push and should never even have started the porting process.  
The OL2 base is in much, much better shape now. The previous OL4 porting attempt results have lost their value.

PR #986, opened 2018-07-13
Feature/wmts bundle

LazerTiberius:


AxxL:
@LazerTiberius Another reviewer for the content?

werrolf:
In short: not like this. 
Let's continue the lively discussion in #989, which contains the same middle bulk of commits as this one.

PR #988, opened 2018-07-16
Added new transport service.

rpour:
The transport services uses following parameters from `parameters.yml`.
* ows_proxy3_host
* ows_proxy3_port
* ows_proxy3_connecttimeout
* ows_proxy3_timeout
* ows_proxy3_user
* ows_proxy3_password
* ows_proxy3_noproxy

Usage...

```php
public function testAction()
{
    $response = $this->get('transport')->handle(
        Request::create('https://www.wheregroup.com')
    );

    return new Response($response->getContent());
}
```
Catch Errors.

```php
public function testAction()
{
    try {
        $response = $this->get('transport')->handle(
            Request::create('https://www.wheregroup.com')
        );

        return new Response($response->getContent());
    } catch(TransportException $e) {
        return new Response($e->getMessage(), 500);
    }
}
```

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Transport.php
+++ b/src/Mapbender/CoreBundle/Component/Transport.php
@@ -9,7 +9,7 @@ use Symfony\Component\HttpFoundation\Response;
  * Class Transport
  * @package Mapbender\CoreBundle\Component
  */
-class Transport
+class Transport implements TransportInterface
 {
     protected $proxyHost = '';
     protected $proxyPort = '';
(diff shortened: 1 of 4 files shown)


werrolf:
As already discussed with @rpour in person, due to considerations arising from recent project developments we will not go forward with the initial plan of anchoring this directly into Mapbender for central availability. Instead, we will opt for a new module that can be shared between multiple projects, including Mapbender.

I am closing this pull now. When the new module's repository is off the ground, I intend to drop a link here for transparency.

PR #989, opened 2018-07-17
Add WMTS feature!

Genc977:


werrolf:
This branch includes completely inacceptable unrelated changes, like the deletion of several classes that have nothing to do with sources in the slightest, extensive redundant reformatting, reintroduction of long redundant clear containers in twigs, revival of EntityHandlers etc.  

This seems to me like a big old git add . of a working copy overwritten with some ancient version. This is not how you make feature branches.

I'm leaving this branch dangling in case somebody has an active working project that requires it, but this will never, ever be merged.

CharlyUnicorn:
I have a running project that requires it :earth_asia:

werrolf:
Actually, I love me some WMTS. I've just had another discussion today that I very much want WMTS support to be a standardized Mapbender thing.

But this somehow manages to undo a year or more of development in certain areas of Mapbender. E.g. why would the introduction of support for a new map service protocol delete a bunch of files in bundles not even related to map service protocol? This is not how this works. 

I can't even tell how this works, if it works. 5701 sloc of changes is as unreviewable as it gets. Even so, the very obvious number of deletions and changes in logically unrelated parts tells me this isn't the solution I want.

One other curious thing that pops out is that commits on this pull first introduce the WmtsBundle folder, and then a later commit removes it again. What ... how?

PR #1001, opened 2018-08-06
Revert keyword data type back to string Fix #1000

### LINKED ISSUE #1000: Incompatibilities due to keywordlist length fix and column format
Ticket  https://github.com/mapbender/mapbender/issues/705 led to an incompatibility with Oracle databases where the keywordlist itself gets a CLOB after the change. The CLOB makes trouble in some of the requests to the database, like deleting a data source.

So we have to take it back. Maybe cut off the keyword when longer than 256 characters.

I don't know a word this length and the example in #705 is a misconfigured WMS in my view. Since the keywords are not really used in MB, we can cut a word after this long length.

Needs to fallback to 3.0.5

DeEgge:
Changed keyword data type back to string. If a source is configured to have a longer keyword list, it will now be cut off after 255 chars.

### DIFF
--- a/src/Mapbender/CoreBundle/Entity/Keyword.php
+++ b/src/Mapbender/CoreBundle/Entity/Keyword.php
@@ -27,7 +27,7 @@ abstract class Keyword
 
     /**
      * @var string $title The source title
-     * @ORM\Column(type="text", nullable=false)
+     * @ORM\Column(type="string", nullable=false)
      */
     protected $value;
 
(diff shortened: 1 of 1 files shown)


AxxL:
Should be also done with 3.0.5

werrolf:
This is not safe for multibyte strings. Should use mb_strlen and mb_substring to avoid e.g. splitting a UTF-8 codepoint in two, which would lead to a corrupt string.  

Please always use proper blocks (curly brackets, body starts a new line) for your ifs.

Issue #632, opened 2017-09-13
backend Application - layerset - choose datasource - resize dialog - make it bigger


astroidex:
* the dialog to chosse a WMS for your layerset is very small - height:400px  witdh:500px
* would be great to have it bigger for usebility
* suggestion height: 700px and width: 700px

.popup {
    min-width: 300px;
    min-height: 250px;
    position: relative;
    z-index: 9999;
    width: 500px;
....
}
![chosse_datasource_small](https://user-images.githubusercontent.com/996298/30368265-f96f7eee-9870-11e7-9882-6d678ff97ac0.png)
![chosse_datasource_big](https://user-images.githubusercontent.com/996298/30368276-01f26d10-9871-11e7-83ad-cfa524911616.png)

LazerTiberius:
Is there any reason why we should use 700px. 700 seems to be a bit random. I would suggest to just open a full-screen overlay there. There is no possibility to interact with something other than the open pop up. 
 But I´m not sure which other elements use the CSS class .popup.

astroidex:
* 700x700 is random. Full--screen would be ok for me too. Only something bigger and easy to handle to improve the usebility

astroidex:
@axxl @werrolf @LazerTiberius - please have a look again. Would enhance the usabiility.

inacht:
popup dialog height and wide resized to 700 pixel
See PR https://github.com/mapbender/mapbender/pull/1004

astroidex:
I tested. Please put height to 500 and change width to 700. As on smaller screens else you will not be able to see the ok button at the bottom

inacht:
@astroidex ok, changed height to 500, and width to 700 pixel.

PR #1004, opened 2018-08-06
backend Application - layerset - choose datasource - resize dialog 

inacht:
resized popup dialog height and wide resized to 700 pixel
Ticket https://github.com/mapbender/mapbender/issues/632


werrolf:
We don't accept updates that replace old arbitrary hardcoded values with new arbitrary hardcoded values.

Issue #990, opened 2018-07-18
Browser support in 3.0.7.3 - Internet Explorer 


CharlyUnicorn:
in the new version 3.0.7 I created the following error behavior:

In the previous versions IE could be used.
In the new version, however, the simfony strip is very big and the application throws an error. You can't use any application or element. 

```
SCRIPT1003: Expected ':'
File: js, Line: 6332, Column: 22
```

![auswahl_054](https://user-images.githubusercontent.com/13517141/42880009-f41bb3f8-8a92-11e8-9695-3466b1bd284d.png)

symfony stripe in app_dev.php developer mode

![ie11 - win8 1 wird ausgefuhrt - oracle vm virtualbox_053](https://user-images.githubusercontent.com/13517141/42879925-a554699a-8a92-11e8-90f7-9ad9494482b2.png)

astroidex:
In some projects we need IE11. Is there a way to support it with 3.0.7.3

AxxL:
If anyone needs support for an outdated webbrowser, although many alternatives exist and are way more popular and widespread, we can build it in.

I bet it is easier to use a different webbrowser like Firefox or Chrome or Safari or even Edge.

CharlyUnicorn:
Sometimes you don't have the possibility to use another browser due to external restrictions. if it is known that the version or browser itself is not supported, then of course you can control it - 
i din't knew it and perhaps you could add this info in the release note from 3.0.7.3?

LazerTiberius:
This is a bug because of a missing function statement. @werrolf will fix this asap.

astroidex:
* @LazerTiberius thanks for the feedback. We should make a bugfix release then.
https://msdn.microsoft.com/query/dev12.query?appId=Dev12IDEF1&l=DE-DE&k=k(VS.WebClient.Help.SCRIPT1003)

astroidex:
Fix via in file mapbender/mapbender/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js line 179

`        _updateSource: function($sourceContainer) {`

astroidex:
* tested with IE 11
* creted PR 1005 https://github.com/mapbender/mapbender/pull/1005
* https://github.com/mapbender/mapbender/tree/hotfix/function-definition-ie-browser-bug-990

werrolf:
Fix merged to release/3.0.7, master, ol4.  
There is no tagged release including the fix yet.  
[...]

PR #1005, opened 2018-08-08
function definition fixed with : see #990

astroidex:
* fixes problem with internet explorer browser see https://github.com/mapbender/mapbender/issues/990
* tested with IE 11


werrolf:
Wrong start branch.
[1008 is equivalent](https://github.com/mapbender/mapbender/pull/1008) and has the correct start branch.

astroidex:
ok. @werrolf I will ask you about start branch

PR #1012, opened 2018-08-29
Internalize mapbender versioning into mapbender repository

werrolf:
Branding, the everlasting joy of life.

This provides Mapbender-internal Mapbender versioning and naming, while still supporting explicit project branding overides.

A new compiler pass displaces [fom: server_version et al](https://github.com/mapbender/mapbender-starter/blob/644dbc24ece67d04bdd9dedcb8baf413d9cc5021/application/app/config/parameters.yml.dist#L20) as the authoritive source of the Mapbender version. Mapbender [gains version knowledge inside itself](https://github.com/mapbender/mapbender/blob/4e86c9f387ac9ee7e3e7737275c9fac29a8d5bce/src/Mapbender/CoreBundle/Resources/config/mapbender.yml#L20).

[Any `server_version` or `server_name` values configured in FOM that follow certain well-treaded patterns are dropped on the floor and replaced by Mapbender's own](https://github.com/mapbender/mapbender/blob/4e86c9f387ac9ee7e3e7737275c9fac29a8d5bce/src/Mapbender/CoreBundle/DependencyInjection/Compiler/ProvideBrandingPass.php#L80). If they deviate significantly from historic default values used in vanilla Github versions of Mapbender Starter, we treat them as desired project-specific overrides and leave them alone.

We also add new parameter keys for [direct project-specific overrides of name, version and logo](https://github.com/mapbender/mapbender/blob/4e86c9f387ac9ee7e3e7737275c9fac29a8d5bce/src/Mapbender/CoreBundle/DependencyInjection/Compiler/ProvideBrandingPass.php#L50). These will always win out, even if they happen to say "Mapbender3" "v3.0pre2" (which would be ignored if it appeared under fom : server_*).

Motivation: currently, git tags are the only hard source of Mapbender versioning information inside the actual Mapbender repository. `fom`: `server_*` is defined in another repository. This mandates parallel updates of multiple repositores at every versioning cycle and throws a wrench into introspection.

This pull obsoletes Mapbender version information in Mapbender Starter.

### DIFF
--- a/src/Mapbender/CoreBundle/Command/VersionCommand.php
+++ b/src/Mapbender/CoreBundle/Command/VersionCommand.php
@@ -1,0 +1,52 @@
+<?php
+
+
+namespace Mapbender\CoreBundle\Command;
+
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class VersionCommand extends ContainerAwareCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('mapbender:version')
+            ->setHelp('Display Mapbender version')
+            // --version is taken by Symfony...
+            ->addOption('--number-only', null, InputOption::VALUE_NONE, 'Display only version (default: name and version)')
+            ->addOption('--name-only', null, InputOption::VALUE_NONE, 'Display only name (default: name and version)')
+            ->addOption('--project', null, InputOption::VALUE_NONE, 'Display project [name and] version instead of Mapbender [name and] version')
+        ;
+    }
+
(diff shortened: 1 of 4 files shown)


werrolf:
We now also get a console command mapbender:version. This is intended mainly to feed into outside scripts that require a version for packaging or other tasks, such as [ComposerBootstrap::genApiDocumentation](https://github.com/mapbender/mapbender-starter/blob/644dbc24ece67d04bdd9dedcb8baf413d9cc5021/application/src/ComposerBootstrap.php#L250).

With the following in my parameters.yml:
```yml
parameters:

    branding.project_name: Awesomebender
    branding.project_version: 9000
```

I receive the following output:
```sh
application$ app/console mapbender:version
Mapbender 3.0.7.4
application$ app/console mapbender:version --project
Awesomebender 9000
```

Likewise, my backend now displays `v. 9000` top-right, and my unmodified AboutDialog welcomes me to `Awesomebender Version 9000`.

PR #1036, opened 2018-10-18
no * for optional fields that are required: false

astroidex:
* do not show the * after the label of a text field (optional field) in the print form if required: false is set

comment:
  label: comment
  options:
    required: false


werrolf:
Merged into release/3.0.7 as of fbb58414d9ff40976826d9a1302c62951a05d567.  
Github will no longer allow changing base to reflect this.

PR #1045, opened 2018-11-07
WMSLoader select children

simonseyock:


simonseyock:
The `useDeclarative` property was not respected anymore in the current mapbender. This is fixed with commit e227f44

astroidex:
I tested the branch. Works like expected. Please merge to the next release.

astroidex:
Hi @simonseyock,
hab das Feature getestet. Läuft alles wie erwartet.

PR kann in die nächste Version rein: 
* https://github.com/mapbender/mapbender/pull/1045

werrolf:
Merged 0544517039be4b13f9bb9a79ecedd70a314d097f, throwing away the rest.  
Apparently, that commit makes it so the root layer of the newly added WMS starts out selected in the Layer Tree (by not getting its treeOptions.selected reset to false).

PR #1047, opened 2018-11-08
Fix/clean up vendorspecifics

werrolf:
Separates VendorSpecific from Dimension inheritance chain.  
Removes unevaluated, functionless attributes extent / origextent / origextentextent (srsly) and `type` (single / interval / multiple) from VendorSpecific class, along with the related backend form fields and processing.

Cherry-picks 467ed5466c755467b51983349686eb828465ec4e from feature/refactorDimension and continues removal of JOII dependencies from mapbender.wms.dimension.js JavaScript classes.


werrolf:
Backref PR #1048

PR #1049, opened 2018-11-09
Fix/dimensionhandler backend

werrolf:
Fixed: when adding a new "DimensionSet" to a DimensionsHandler Element (click on `+` icon), selecting an instance now properly initializes the extent slider, so the "DimensionSet" can immediately be configured.  
Fixed: the Element properly controls multiple sources.  
Fixed: the Element can control sources from different Layersets.  
Collateral: the markup for a DimensionsHandler rendered into a toolbar (top or bottom) has been converted to Bootstrap 3 to avoid `position: absolute` (Mapbender responsive push). Looks identical with Mapbender's default CSS.

Remaining known issues:
New DimensionsHandler Element can still not be configured properly before saving once. Full initialization of the form requires the DimensionsHandler to be saved once with the Map target has been selected.   
A "DimensionSet", once configured, cannot be deactivated by deselecting the last controlled source; the backend slider remains active with the previous valid extent; the DimensionSet has to be deleted (`x` icon).  
"Open ended" or "incomplete" (though fully valid) Dimension specs such as "2016-01/2018-11" still throw errors. Only Wms dimensions with explicit start + end + resolution ("2016-01/2018-11/P1M") are fully supported.  
Backend form does not show required fields and does not show error messages when required fields are missing. The backend form only saves successfully if each "DimensionSet" has a title, extent range, and controls at least one source.

This branch further builds on the [fix/clean-up-vendorspecifics](https://github.com/mapbender/mapbender/pull/1047) branch, so some commits listed here are not specifically dealing with DimensionsHandler.


CharlyUnicorn:
excellent work, thanks!
I tested the changes in a project: the wms-t are now much more pleasant to integrate and the application is more intuitive to use.
Through the improvements, thematic layersets can now be used and the integration of several wmt-s via a central slider is possible. Furthermore this eliminates the problems with the checkbox and the use of the small slider in the layertree.

dim_handler

PR #1057, opened 2018-11-26
fix MobileTemplate FeatureInfo scrollable

astroidex:
* Reported bug from David Müller via Mailing-list
* https://lists.osgeo.org/pipermail/mapbender_users/2018-November/005522.html


werrolf:
LGTM

PR #1058, opened 2018-12-06
todo solved, parameter offset is interpreted and considered

frwg:


werrolf:
We don't want mutable popups going forward. This was a 7 year old todo that didn't make much sense to begin with, but got copied when the code came over from FOM.

PR #1065, opened 2019-01-04
Change the font-awesome path for font awesome version 5

GeOSMRob:
After the new font-awesome version has been released via "mapbender / vis-ui.js", the path for font-awesome in the icons.scss must be adjusted. After this adjustment the new version 5 will be loaded.

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/sass/libs/_icons.scss
+++ b/src/Mapbender/CoreBundle/Resources/public/sass/libs/_icons.scss
@@ -1,4 +1,4 @@
-@import url("components/font-awesome/css/font-awesome.min.css");
+@import url("components/font-awesome/font-awesome-built.css");
 [class^="icon"],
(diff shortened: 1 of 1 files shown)


werrolf:
No, the new version 5 will not be loaded after this change.  
The built css is generated from all declared css sources in the vendor/components/font-awesome/css folder. In 5.x this happens to include the v4-shims.css, which partially restores compatibility.  

Mapbender is currently _incompatible_ with FA 5.*, with or without the shims.

That said, the proposed change is compatible with FA 4.* as well as with FA 5.*. So I'm merging it.  
But do note that if you want to punch FA 5.* into a Mapbender installation at your own risk, you will have to change the [^4.0 requirement in your Starter](https://github.com/mapbender/mapbender-starter/blob/bcea11c082792512c36ea911e56cf3731351bd48/application/composer.json#L62).

PR #1066, opened 2019-01-11
[RFC] Revert Template class magic

werrolf:
Reverts a bunch of "Refactor and improve" that ended up turning the Template classes into [a mess of self-modifying static attributes](https://github.com/mapbender/mapbender/blob/v3.0.7.7/src/Mapbender/CoreBundle/Component/Template.php#L224).

The methods getAssets, getTitle are the only ones that are ever called on a Template from the outside. This has been true since years before the "refactor and improve" period, and still remains true.

Doing this with methods allows standard inheritance. An inheriting template's getAssets('js') could simply do
```php
return array_merge(parent::getAssets('js'), array(
   'something-extra.js',
));
```
This would just work, enabling independent maintenance on the base class asset dependencies without side effects in a customized project Template.

Static properties can't do this.  
Placing placeholder fallback logic at the abstract Template level prevents language-level enforcement of the API, because every bit involved in this must have a non-abstract implementation, as we see with the current code.   
Self-modification through the hierarchy to circumvent limitations of static properties is a dirty hack working around the initial dirty hack.

We are already heavily impeded in our ability to remove obsolete libraries or replace them with newer versions, and the way Template currently works multiplies this issue across projects on the most fundamental level of project is customization.

This pull looks like a straightforward change only affecting the Template hierarchy itself, because that's exactly what it is.

Unfortunately, this has very real compatibility impact on customized project-level Template children implemented in the "refactor and improve" fashion. The `listAssets` method, and nearly all the static properties, lose their meaning. The template will silently not do anything.

No solution has been found to remain backwards-compatible with "refactor and improve"-style custom Template classes that would not, again, require new built-in magic, hacks and workarounds.

### DIFF
--- a/src/Mapbender/CoreBundle/Component/Application/Template/IApplicationTemplateAssetDependencyInterface.php
+++ b/src/Mapbender/CoreBundle/Component/Application/Template/IApplicationTemplateAssetDependencyInterface.php
@@ -1,0 +1,40 @@
+<?php
+
+
+namespace Mapbender\CoreBundle\Component\Application\Template;
+
+
+interface IApplicationTemplateAssetDependencyInterface
+{
+    /**
+     * Should return the list of assets of given $type that are required for the template to function.
+     *
+     * Supportable $type values are 'css', 'js', 'trans'.
+     *
+     * 'js' and 'css' entries should be either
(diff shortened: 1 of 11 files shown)


werrolf:
Closing because clearly not safe to merge at this point. Discussions welcome.

PR #1075, opened 2019-02-01
Change Twig url function to path for compability with reverse proxy installations

LazerTiberius:
If you use Mapbender behind a reverse proxy url-funciton will generate links to the internal server no to the reverse proxy.


werrolf:
Yes.

PR #1076, opened 2019-02-01
Fix broken applications with undefined click configuration

LazerTiberius:
In twig "if(variable)" only checks if the value has a non null value but throws an error if the variable is not present, so we need to check this first.


werrolf:
YAML applications are special snowflakes indeed.

PR #1080, opened 2019-02-06
Enh/improved line pattern quality in export + print

werrolf:
Polygonal features generated by Digitizer or similar developments can have quite complex line styling, which has historically not transferred very well into ImageExport / Print.

This pull improves support for high line widths, non-trivial line opacities and line pattern styling. This is achieved by
1) Compositing. The polygon and its outline are both prepared on a temporary sub-canvas, where holes can be cut without undesired removal of base image portions, and where blending artifacts can be avoided by simply turning blending off.
2) Deconstructing the line pattern into individual line and point primitives.

Consider geometries styled like this (browser view, Digitizer):
poly-browser-view

The best you could previously achieve in a Mapbender printout ended up something like this:
poly-badness

Things of note:
* The outline opacity escalates around steep edges, where the thick outline self-intersects and incrementally blends its color onto the background several times
* The "donut" hole is not cut out, but filled with the polygon color "normally"
* The line pattern comes out completely jumbled
* Sharp angles produce very abrupt changes between completely horizontal and completely vertical line segment caps produced by GD. This is especially obvious for the the lone purple dot, which was created as a tiny circle geometry. It comes out barely recognizable as a cross.

Result after changes in this pull:
poly-improvements

The donut cutout works.  
The outline pattern, while admittedly still not 100% perfect, is much smoother.  
Intermittent vertices are rounded out.  
The tiny circle geometry looks much closer to what it should be.

Sidenote: I'd also like to document the recent evolution of point rendering here
point-evo
v3.0.7.7 could not render a point outline with any other width than 1 pixel, at all, following an unfortunate GD limitation.  
The middle image is an attempt to resolve this by creating polygonal geometry for the point outline.  
The right version shows the current development version, which renders the outline as a fully transparent circle stamped out of a non-transparent circle, which then goes through compositing. The result is indistinguishable from the browser view.

### DIFF
--- a/src/Mapbender/PrintBundle/Component/Geometry/LineLoopIterator.php
+++ b/src/Mapbender/PrintBundle/Component/Geometry/LineLoopIterator.php
@@ -1,0 +1,32 @@
+<?php
+
+
+namespace Mapbender\PrintBundle\Component\Geometry;
+
+
+class LineLoopIterator extends LineStringIterator
+{
+    public function __construct(array $points)
+    {
+        if (count($points) < 2) {
+            throw new \InvalidArgumentException("Cannot form a loop from " . count($points) . " points");
+        }
+        parent::__construct($points);
+    }
+
+    public function valid()
+    {
+        // unline parent, allow iteration to continue even if lineEndIterator is one past the last point
+        return $this->lineStartIterator->valid();
+    }
+
+    public function current()
+    {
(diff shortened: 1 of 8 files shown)


werrolf:
Fuzziness could not be resolved within a reasonable timeframe. Postponing further work.

Issue #1082, opened 2019-02-06
[map][visibleLayers] root layer can't get activated by visiblelayer parameter on start & toggle is gone


astroidex:
**Describe the bug**
Documentation see: https://doc.mapbender.org/en/functions/basic/map.html#make-layer-visible

When you have a service deactivated on start (but the layer active) you can not use visiblelayers to activate the root layer. 

https://demo.mapbender.org/application/mapbender_digitize_demo?visiblelayers=50/150,50/151,50/152

**To Reproduce**
1. https://demo.mapbender.org/application/mapbender_digitize_demo -> Mapbender user root layer is disabled, the other layer are active, toggle is on
2. https://demo.mapbender.org/application/mapbender_digitize_demo?visiblelayers=50/150 -> 50/150 is the rot layer. Request should activate the root layer and like this activate all the layers oft the Service (as there are on by configuration) It is still not activated and the toggle is gone (service is closed)
3. https://demo.mapbender.org/application/mapbender_digitize_demo?visiblelayers=50/150,50/151,50/152 -> root layer activated butt toggle is gone

**Expected behavior**
* see above

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.0.7.7.

**Screenshots / Logmessage**

**Additional context**
With the following SQL you can create the requests for all WMSInstances:

`
Select slug, title,
'https://demo.mapbender.org/application/'
|| foo2.slug ||  '?visiblelayers=' || string_agg(combi, ',') as
mb_visibleLayerId
from
(SELECT li.id, li.title as title_Layerinstance, wmsinstance || '/' || li.id as combi, parent, wmsinstance, wmslayersource, foo.application_id, foo.slug, foo.wms_instance_id, foo.layerset, foo.title , foo.layerset_title
, 'https://demo.mapbender.org/application/' || foo.slug || '?visiblelayers=' || wmsinstance || '/' || li.id as mb_visibleLayerId
--parent, li.title, active, allowselected, selected, info, allowinfo, toggle, allowtoggle, allowreorder, minscale, maxscale, style, priority
          FROM public.mb_wms_wmsinstancelayer li,
                                ( SELECT l.id as layerset_id, l.application_id, l.title as layerset_title,
  s.id as wms_instance_id, layerset, s.title, weight, enabled, discr, basesource, a.slug
   FROM public.mb_core_layerset l,
  public.mb_core_sourceinstance s,
  public.mb_core_application a
  where s.layerset = l.id
[...]

PR #1083, opened 2019-02-09
Fix/declarative mb wms layers repeatability + empty group layer names

werrolf:
Fixes non-repeatable behaviour in handling of 'declarative' links that dynamically add or update Wms sources with explicit layers via `mb-wms-layers` attribute _and_ (default) `mb-wms-merge` "1". Previously, the performed operations on the source were different for newly added sources and already loaded sources.  Now a repeated click on the same declarative link performs the same incremental (`mb-layer-merge` omitted or "1") or total (`mb-layer-merge` "0") set of changes to the source.  

Supports `mb-layer-merge` attribute on 'declarative' links as documented. Previously the code respected only the `mb-wms-layer-merge` attribute, which wasn't documented. Now it supports both equivalently.  

Enhanced support for activating layers explicitly even on sources with intermediate group or root layers with empty names. Previously such layers could not be acted upon distinctly, because `mb-wms-layers` can only identify layers with distinct names. Now the 'declarative' `mb-wms-layers` interaction implicitly activates also the parents of the named layers (including the root layers). This supports all relevant workflows nicely. It also increases robustness against structural changes of the WMS layer setup (regrouping, increase of nesting).

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
@@ -637,7 +637,7 @@
                         options.layers[$layerLi.attr('data-id')] = value;
                     }
                 });
-                self.model.changeLayerState(source, options, false, true);
+                self.model.changeLayerState(source, options, null);
             });
             return false;
         },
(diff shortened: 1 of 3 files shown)


werrolf:
The introduced cleanups to geosource changeLayerSource appears to also collaterally have fixed #1082.

PR #1088, opened 2019-02-12
Fix/redlining sidepane deactivate tools

### LINKED ISSUE #992: Redlining in 3.0.7 - error to deactivate in sidepane and missing display in map
in the new version 3.0.7 I created the following error behavior: 

If the element is loaded into the sidebar, the sketch functions are never deactivated when the checkbox "deactivate after closing" is not active and no further functions can be used after the starf of a sketch. If the checkbox "deactivate after closing" is activated, then you can not activate / use the tool after the first drawing action.

If the element is included as a dialog, then the sketch functions works until the window is closed. If, after closing the dialog window, the module is activated again via the button, the elements no longer appear in the map. The list in the dialog will continue to list all items, but you don’t see them.

werrolf:
Piggy-backs on #1087, which introduce visibility information machinery for sidepane-dwelling element widgets.  
Fixes #992  

Redlining in sidepane is rebased to inherit from mbBaseWidget, and implements the new `hide` method to deactivate its drawing control.


astroidex:
Tested works fine

Issue #1050, opened 2018-11-13
Buttoninteraction is not transparent


astroidex:
**Describe the bug**
Buttonhandling is not consistent. Reported by OBK

**To Reproduce**
1. tbd

**Expected behavior**
A clear behaviour of the buttons. 
* close via x-Button in dialog should deactivate the button (except featureInfo - when deactivate on close=false)
* close-Button at bottom of dialog should deactivate the button 
* click on button in toolbar should deactivate the button
* if button is in a group: click on another button of the group should deactivate the button

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.0.7.7

werrolf:
As we've discussed before, it's not the buttons that are inconsistent, it's the elements that are inconsistent. The external control API to invoke on an Element to get it to cooperate is highly variable. This necessitates deep knowledge, or glances into the code, to correctly configure a controlling button.

Github won't let me attach the spreadsheet here unfortunately, but I've done some digging.

There's one big family that supports `open` and `close`. Another smaller family that supports `activate` and `deactivate` (Redlining, Ruler _open_ _a_ _popup_ on activate). Some support both interchangeably (Redlining, Coordinates Utility). One (POI) only supports the weird pair of `activate` and `close`.
Not all of these are intended to be controlled by an outside button, but they are capable. E.g. AboutDialog and GpsPosition both act quasi as their own button, but their APIs also work fine for Button control.

Some Elements only ever work in a sidepane (Digitizer), while others only ever work in a popup (POI, SuggestMap and friends).

It's these inconsistencies that need to be tackled before we can have straightforward button configurations.

werrolf:
Also, there is really no way to tell for a button whether or not its controlled Element is active or not. Elements have multiple, inconsistent ways to "open automatically", so not even the initial state for the button is known.

Also see #549.

rschwammborn:
In Mapbender 3.0.8, buttons in the default toolbar perform as described under "Expected Behaviour".
Closing this issue for the moment, feel free to reopen if new button complexities appear.

Issue #549, opened 2017-06-22
Button - Highlight and selection of POI-button in group


CharlyUnicorn:
You can add buttons in the toolbar of an application. You can group these buttons by the function "group". This function adds the element to a group. Only one button of the group may be activated and these button is highlighted in a different color. 
If you close the window of the selected button, then the highlighted button is deactivated and the special coloring should disappear. 

This function works very well except for the button of the POI-function. 
If you add a button for the poi-funktion, add it to the group and activate this button, then these button is highlighted in a different color. But when the button is disabled, the color highlighting is still active until you click another button. 

![auswahl_028](https://user-images.githubusercontent.com/13517141/27431088-79a333a6-574b-11e7-9b96-5b6304085d5c.png)

You can test this: 
- https://demo.mapbender3.org/application/buttons_group
- Click on a button and close the popoup via the "close" button
- should work fine exept for the poi-button

AxxL:
@CharlyUnicorn Was willst Du denn mit den Gruppen erreichen?
Soweit ich sehe, hat die Legende das gleiche Problem.

werrolf:
Auto-closed by Github commit message parsing. Fixed in [ea18664](https://github.com/mapbender/mapbender/commit/ea18664fc2e618ae77de3233376896a207cb754f).

PR #1089, opened 2019-02-13
[WIP] Enh/remove button activate deactivate methods poc

werrolf:
Proof of concept version of [enh/grand-unified-button-api](https://github.com/mapbender/mapbender/tree/enh/grand-unified-button-api): empties any 'activate' and 'deactivate' options the Button receives from the Backend (or a YAML application), and thus forces it to figure it out all on its own.

Observed behaviour:  
Button auto-detects a probable initial activation state of its target widget by evaluating [various](https://github.com/mapbender/mapbender/commit/04b50b5b392106d817ca49847a365bb821769565#diff-47bf66afd39ec612d69958ee81aafcb8R176) configuration options.

Button can activate and deactivate all but one of the tested widgets properly. Button self-assigns an adequate highlight (without needing the single-member 'group' setting workaround). Note: button will still deactivate other members of the same group. Other non-grouped targets, or targets in a different group, will remain active.

The only outlier that didn't work fully in testing is Coordinates Utility, which does not close on the second click, and can have inverted button highlighting states.

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.button.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.button.js
@@ -15,7 +15,7 @@
         $toolBarItem: null,
         /** initialized to true if button can determine target's active state and remember its own active state */
         stateful: null,
-        actionMethods: {},
+        actionMethods: null,
 
         _create: function () {
             if (this.options.click) {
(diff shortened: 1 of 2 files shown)


werrolf:
This should resolve #1050 and #549. It also should make it easier to set up buttons that work, because there are now two less options to think about.

werrolf:
No longer relevant.

PR #1094, opened 2019-02-15
Enh/reverse proxy compatibility

LazerTiberius:
Changed all twig path functions to  url because in specific reverse proxy environments url() generates URLs based on the host and not the reverse proxy. So some links will not be accessible if the application is opened through the proxy.


werrolf:
You can do whatever you need on tmp/90001, but it's not an acceptable merge origin.

LazerTiberius:
> You can do whatever you need on tmp/90001, but it's not an acceptable merge origin.

Already closed because of this. Accidentally based this branch on tmp/90001.

PR #1095, opened 2019-02-15
Fix/button auto action method detection

werrolf:
This makes Buttons controlling other Elements work even if the backend `activate` / `deactivate` configuration is completely empy, or kind of bad. The "correct" settings for these options can be [non-intuitive](https://github.com/mapbender/mapbender/issues/1050#issuecomment-463119829) and figuring that out is better left to [a piece of code](https://github.com/mapbender/mapbender/commit/d8be3b5dc64decfe752f054c45a99ceecda52ab4#diff-47bf66afd39ec612d69958ee81aafcb8R90).  

NOTE: if there are any backend `activate` / `deactivate` settings they are respected insofar as they are the highest-priority picks of methods to use to control the target element _but_ _only_ _if_ the target widget has such a method. This retains administrative control and the option to call entirely non-standard, randomly named methods.  
OTOH it auto-heals any erroneous configurations by finding something that works. So the straightforward approach of leaving those backend values completely empty, will now just work.

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.button.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.button.js
@@ -13,9 +13,7 @@
         active: false,
         targetWidget: null,
         $toolBarItem: null,
-        /** initialized to true if button can determine target's active state and remember its own active state */
-        stateful: null,
-        actionMethods: {},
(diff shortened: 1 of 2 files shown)


werrolf:
This should resolve #1050.

Issue #1104, opened 2019-02-21
[PrintClient] Replace Pattern does not work


GeOSMRob:
**Describe the bug**
If I set the Replace Pattern in the print element, then only the "default" is evaluated and replacement is not evaluated.

**To Reproduce**
1. Go to PrintClient
2. Click on PrintClient-Configuration
3. Scroll down to Replace pattern
4. Set replace_pattern-Params
4. See error to Printing

**Docs**
- https://doc.mapbender.org/de/functions/export/printclient.html#yaml-definition

**Versions:**
 - Mapbender: [3.0.7.7]
 - Browser [Chrome,Firefox]

**Code:**
- https://github.com/mapbender/mapbender/blob/release/3.0.7/src/Mapbender/PrintBundle/Element/PrintClient.php#L367

GeOSMRob:
Problem was found and a solution is worked out.

werrolf:
See PR#1105

werrolf:
We're good now right?

GeOSMRob:
Alright bug has been fixed and the replace pattern works now.

PR #1105, opened 2019-02-21
fix addReplacePattern Function

GeOSMRob:
Now you can use different and multiple replacement patterns.


werrolf:
LGTM.

werrolf:
See Issue #1104

PR #1116, opened 2019-03-08
Feature/wmts homecoming

werrolf:
Reintegrates [the external Wmts bundle](https://github.com/mapbender/wmts) with Mapbender to allow ongoing architectural work on source and source instance handling to proceed in a unified fashion.

Side-by-side installation of the external bundle with this branch is not possible. This is a full displacement, using the same PHP namespace. An appropriate package conflict rule will be added to prevent simultaneous installation.   
Some of the documented limitations and quirks of the external bundle have been resolved, but enough remain that this is still an experimental feature that we cannot reasonably enable by default.   

### Known limitations / quirks
* Only a single layer can be enabled per WMTS / TMS instance. To get different layers from the source to display, additional instances need to be added to the same application
* WMTS / TMS sources cannot be "opened" in the Layertree, they are either on or off with no further control possible
* There is currently no support for Metadata via Layertree sub-menu.
* There is currently no support for FeatureInfo
* There is currently no support for vendorspecifics or dimensions
* What appears as the root layer in the instance configuration will fail to save its various checkbox settings, most notably the initially enabled state of the source. This means that WMTS / TMS instances added to an application will always start out enabled.
* WMTS / TMS sources will magically self-disable when switching to an incompatible SRS. Layertree checkboxes will uncheck themselves. Depending on current SRS, Layertree visual may become inconsistent when you click on the source
* There is no integration with any WmcBundle functionality (SuggestMap, WmcLoader, WmcEditor et al). Having active WMTS or TMS instances in the same application as these Elements may lead to incomplete states getting saved, generation of states that are incompatible with other applications, and other errors.
* There is no mechanism to add WMTS or TMS sources to YAML-defined Applications. Only database Applications can have them.
* WMTS and TMS sources cannot be reloaded. To effectively "update" them, you will need to delete it, load it as a new source, and create new instances.

### Resolved limitations (vs [external bundle documentation](https://github.com/mapbender/wmts/blob/master/README.md))
* There are no restrictions on the Map Element's dpi and scales settings. As with any fixed raster source, do note however that going off the "ideal" values may induce tile rescaling, and thus reduce visual quality.
* There are no restrictions on the Map Element's max extent
* There are no restrictions on the Map Element's initial `SRS` and switchable `Other SRS` setting
* There are no restrictions on the Map Element's tile size setting. Tile sizes for WMTS / TMS are predetermined by the source, are automatically selected as appropriate, and cannot be configured. The Map's tile size setting remains functional for WMS tiling and can be set to any desired value.
* "Zoom to layer" via the Layertree sub-menu is supported
* Current layer opacity is applied when exporting / printing
* Print scale is respected in the same way as for WMS sources

### Enabling
To enable WMTS and TMS source support, even with this branch integrated, two additional steps are required:
1) The bundle must be added to the [application kernel's bundle registration](https://github.com/mapbender/mapbender-starter/blob/v3.0.8-beta1/application/app/AppKernel.php#L11), like so:
```diff
--- a/application/app/AppKernel.php
+++ b/application/app/AppKernel.php
@@ -30,6 +30,7 @@ class AppKernel extends Mapbender\BaseKernel
             // Optional Mapbender bundles
             new Mapbender\WmcBundle\MapbenderWmcBundle(),
             new Mapbender\WmsBundle\MapbenderWmsBundle(),
+            new Mapbender\WmtsBundle\MapbenderWmtsBundle(),
             new Mapbender\ManagerBundle\MapbenderManagerBundle(),
             new Mapbender\PrintBundle\MapbenderPrintBundle(),
             new Mapbender\MobileBundle\MapbenderMobileBundle(),
```
2) The bundle's controller namespace must be added to [the routing configuration](https://github.com/mapbender/mapbender-starter/blob/v3.0.8-beta1/application/app/config/routing.yml), like so:
```diff
--- a/application/app/config/routing.yml
+++ b/application/app/config/routing.yml
@@ -15,6 +15,10 @@ mapbender_wmsbundle:
     resource: "@MapbenderWmsBundle/Controller/"
     type: annotation
 
+mapbender_wmtsbundle:
+    resource: "@MapbenderWmtsBundle/Controller/"
+    type: annotation
+
 mapbender_coordinatesutilitybundle:
     resource: "@MapbenderCoordinatesUtilityBundle/Controller/"
     type: annotation
```


werrolf:
I don't know what's going on with Github today, but this branch has most definitely just been merged.

PR #1119, opened 2019-03-15
Feature/add confirm dialog

frwg:
Move confirmDialog (so fare used in Digitzier only) to Mapbender Core


werrolf:
Data manager and by extension Digitizer confirmation dialog: https://github.com/mapbender/data-manager/blob/master/Resources/public/DialogFactory.js#L35

Doesn't use vis-ui.
Uses a promise instead of callback options.
Has sane, self-provided css rules.

We're not reintroducing vis-ui dependencies, much less into Mapbender core.

Sorry.

Phocacius:
This feature is implemented in the data-manager subrepository in the meantime: https://github.com/mapbender/data-manager/blob/master/Resources/public/DialogFactory.js#L51
This PR is therefore not necessary anymore.

PR #1125, opened 2019-03-22
Enh/yaml app layer groups

DeEgge:
This pull enables grouped layers configureable in a YAML application.

If a WMS was configured to have grouped sublayers (i.e. http://wms2016.wheregroup.lan/cgi-bin/mapserv?map=/data/umn/germany/germany_group.map&REQUEST=GetCapabilities&SERVICE=WMS&VERSION=1.3.0), loading it as a source via the backend would result in correctly grouped layers. This now works in a YAML-based application as well.

image

vs.

```yaml
germany_group:
    class: Mapbender\WmsBundle\Entity\WmsInstance
    title: Germany Group
    url: http://wms2016.wheregroup.lan/cgi-bin/mapserv?map=/data/umn/germany/germany_group.map
    version: 1.3.0
    visible: true
    format: image/png
    transparent: true
    tiled: false
    opacity: 100
    layerorder: standard
    layers:
        11: { name: laender_info, title: laender mit info, visible: true }
        12:
            name: group1
            title: group1 rot
            visible: true
            layers:
                13: { name: g1_l1_Bundeslaender_circle, title: g1 l1 federal states, visible: true }
                14: { name: g1_l2_Bundeslaender_hatch, title: g1 l2 federal states, visible: true }
        15:
            name: group2
            title: group2 gruen
            visible: true
            layers:
                16: { name: g2_l1_Bundeslaender_circle, title: g2 l1 federal states, visible: true }
                17: { name: g2_l2_Bundeslaender_hatch, title: g2 l2 federal states, visible: true }
        18: { name: simple_Bundeslaender, title: alle BL ohne Gruppe, visible: true }
        19:
            name: group3
            title: group3 blau
            visible: true
            layers:
                20: { name: g3_l1_Bundeslaender_circle, title: g3 l1 federal states, visible: true }
                21: { name: g3_l2_Bundeslaender_hatch, title: g3 l2 federal states, visible: true }
                22: { name: g3_l3_Bundeslaender, title: g3 l3 federal states, visible: true }

```

### DIFF
--- a/src/Mapbender/WmsBundle/Component/WmsInstanceEntityHandler.php
+++ b/src/Mapbender/WmsBundle/Component/WmsInstanceEntityHandler.php
@@ -98,7 +98,7 @@ class WmsInstanceEntityHandler extends SourceInstanceEntityHandler
                 ->setName($configuration["name"])
                 ->setTitle($configuration['title'])
                 ->setParent($layersource)
-                ->setId($this->entity->getId() . '_' . $num);
+                ->setId($configuration['title'] . '_' . $num);
             if (isset($configuration["legendurl"])) {
                 $style          = new Style();
                 $style->setName(null);
@@ -118,7 +118,7 @@ class WmsInstanceEntityHandler extends SourceInstanceEntityHandler
             $source->addLayer($childLayerSource);
             $childLayerInstance       = new WmsInstanceLayer();
             $childLayerInstance->setTitle($configuration["title"])
-                ->setId($this->entity->getId() . '_' . $num)
+                ->setId($configuration['title'] . '_' . $num)
                 ->setMinScale(!isset($configuration["minScale"]) ? null : $configuration["minScale"])
                 ->setMaxScale(!isset($configuration["maxScale"]) ? null : $configuration["maxScale"])
                 ->setSelected(!isset($configuration["visible"]) ? false : $configuration["visible"])
--- a/src/Mapbender/WmsBundle/Component/WmsInstanceEntityHandler.php
+++ b/src/Mapbender/WmsBundle/Component/WmsInstanceEntityHandler.php
@@ -85,7 +85,7 @@ class WmsInstanceEntityHandler extends SourceInstanceEntityHandler
         $this->entity->addLayer($rootInstLayer);
 
         foreach ($configuration['layers'] as $layerDef) {
-            $this->populateFromYaml($layerDef, $source, $layersourceroot, $rootInstLayer, $num);
+            $this->populateFromYaml($layerDef, $source, $layersourceroot, $rootInstLayer, $num++);
         }
         return $this->entity;
     }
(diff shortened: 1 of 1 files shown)


src/Mapbender/WmsBundle/Component/WmsInstanceEntityHandler.php
```diff
@@ -79,17 +80,26 @@ public function setParameters(array $configuration = array())
...
-                ->setTitle($layerDef['title'])
-                ->setParent($layersourceroot)
-                ->setId($this->entity->getId() . '_' . $num);
-            if (isset($layerDef["legendurl"])) {
+            $childLayerSource = new WmsLayerSource();
+            $childLayerSource->setSource($source)
+                ->setName($configuration["name"])
+                ->setTitle($configuration['title'])
+                ->setParent($layersource)
+                ->setId($configuration['title'] . '_' . $num);
```
werrolf: This is not safe. Titles can repeat, and sometimes it makes perfect sense for them to repeat. E.g.  
`2018` => `Trees`  
`2018` => `Birds`  
`2019` => `Trees`  
`2019` => `Birds`  
(one might prefer using WMS dimensions here, but you get the point)

If we have to concatenate, I'd prefer concatenating the parent instance layer's id, which is readily available. Sth like
```php
      ->setId($layersource->getId() . '_' . $num)
```

src/Mapbender/WmsBundle/Component/WmsInstanceEntityHandler.php
```diff
@@ -99,30 +109,38 @@ public function setParameters(array $configuration = array())
...
-                ->setMaxScale(!isset($layerDef["maxScale"]) ? null : $layerDef["maxScale"])
-                ->setSelected(!isset($layerDef["visible"]) ? false : $layerDef["visible"])
-                ->setInfo(!isset($layerDef["queryable"]) ? false : $layerDef["queryable"])
-                ->setParent($rootInstLayer)
-                ->setSourceItem($layersource)
+            $layersource->addSublayer($childLayerSource);
+            $source->addLayer($childLayerSource);
+            $childLayerInstance       = new WmsInstanceLayer();
+            $childLayerInstance->setTitle($configuration["title"])
+                ->setId($configuration['title'] . '_' . $num)
```
werrolf: Not safe to use title, see above. Since it has always been equal to the layer source's id, it should be explicitly
```php
    ->setId($childLayerSource->getId())
```
... to avoid any further confusion

werrolf:
Also note that any YAML-defined `layers` mapping I've seen seems to have unique keys in front, including your own example YAML definition. They have simply been ignored so far.

There are two loops in there (one on the root layer creation, one in the recursion) that discard these mapping keys. They could be the ids. Enforcing uniqueness on them is a bit harder though (checking all siblings on all levels), so we'll leave that as is for now.

DeEgge:
Changes were made as requested. Layer IDs now look like this:

```
string(15) "germany_group_0"
string(17) "germany_group_0_1"
string(17) "germany_group_0_2"
string(19) "germany_group_0_2_3"
string(19) "germany_group_0_2_4"
string(17) "germany_group_0_3"
string(19) "germany_group_0_3_4"
string(19) "germany_group_0_3_5"
string(17) "germany_group_0_4"
string(17) "germany_group_0_5"
string(19) "germany_group_0_5_6"
string(19) "germany_group_0_5_7"
string(19) "germany_group_0_5_8"
```

werrolf:
Nice, thanks.

PR #1168, opened 2019-05-16
Fix/allow undefined style

frwg:
Trying to make a deep copy of _undefined_ obviously fails, program should be tolerant here and treat _undefined_ values equal to _null_ values. Additionally, I switched the _if statement_


werrolf:
Too much noise in the pull to salvage.
Label fontOpacity defaults to 1 as of 1fd96e699c814bd45b47f31a8d04bc93e7a405f9 (Feb 28).
Check for non-null style has been adopted as 4d291fe053e52c665fef97883516104047affae8 just now (simple boolean coersion).
Anything else?

werrolf:
If there's no further commenting activity this pull will be closed and the branch deleted on 26th of August.

frwg:
D'accord!

werrolf:
> D'accord!

Ok!

PR #1182, opened 2019-06-15
Feature/simpleSearch simpleSearch srs and label-string configurable

GeOSMRob:
An extension of the simple search:

- Now the objects can be transformed on the client side in projection.

- In the backend, the SRS can now be adjusted for the search

- In addition, it is possible to query Geojson objects via the search

- In addition, the label string in the backend can be adjusted by user-defined settings.

`properties.name + properties.street + properties.housenumber + properties.city + properties.postcode`

Example:

grafik

### DIFF
--- a/src/Mapbender/CoreBundle/Element/SimpleSearch.php
+++ b/src/Mapbender/CoreBundle/Element/SimpleSearch.php
@@ -56,6 +56,7 @@ class SimpleSearch extends Element
             'label_attribute' => 'label',
             'geom_attribute'  => 'geom',
             'geom_format'     => 'WKT',
+            'geom_srs'        => '4326',
             'delay'           => 300,
         );
     }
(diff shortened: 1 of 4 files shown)


astroidex:
Hi @GeOSMRob, is this solutionupdate compatible? At the moment you would refer to the column via "label" without properties.

GeOSMRob:
Hi @astroidex, 
yes this solution is compatible with other search requests.
You can also compile the values from the request in the label field. 

This field is important because, for example, at photon geocoder dynamic resquest parameters exist.

Example from photon request:

- https://github.com/komoot/photon 

Request:
grafik

astroidex:
Hi @GeOSMRob, great! Looks like a great improvement for the SimpleSearch element.

werrolf:
This is a two-feature pull.  
GeoJSON is defined to a fixed CRS. Invalid GeoJSON sources are not Mapbender's problem.  
Your GeoJSON source can and should be configured to emit a document label. This is not Mapbender's problem.

PR #1186, opened 2019-06-24
Revert YamlApp changes

swinkelmann:
Please revert this changes. We cannot run the yaml-apps of two of our customers.
Different reasons:
- Loading parameters from application.yml fails. (parameters 'outside' the application tag)
- Loading parameters from parameters.yml into application configuration also fails.
- View permissions


werrolf:
No.
Application definition Yamls are for applications, not for parameters. This was never even supposed to work.
Put your parameters into parameters.yml or any bundle configuration of your choosing. This gives you inline placeholder expansion of parameters into other parameters as a benefit, and a consistent and controllable order of precedence should a parameter appear in multiple places.

Grants check regression (unsettable VIEW privilege on Yaml applications) has already been fixed independently.

PR #1190, opened 2019-06-26
Replace HTMLHyperlinkElementUtils port property access with RegEx

### LINKED ISSUE #1187: [frontend] add WMS via WMSLoader with InternetExplorer fails
**Describe the bug**
With InternetExplorer 11 WMSLoader fails to load the WMS. Only a message shows up. The same problem occurs for WMS that are loaded with **loadDeclarativeWms**.

**To Reproduce**
1. Open InternetExplorer 11
2. Go to https://demo.mapbender.org/application/mapbender_user_yml
2. Click on **WMSLoader**
3. Load the referred mapbender-user WMS
4. See error **WSLoader: wms capabilities can't be loaded**

With app_dev.php tyou can see that the url is not parsed correct:

`Invalid url 'https://wms.wheregroup.com:443cgi-bin/mapbender_user.xml?SERVICE=WMS&VERSION=1.1.1&REQUEST=getCapabilities'`

**Expected behavior**
Load WMS to application

 - Mapbender: 3.0.8.1 with IE 11 (with firefox, Edge everything works fine)

**Screenshots / Logmessage**
mb_wmsLoader

LazerTiberius:
Replace HTMLHyperlinkElementUtils port property access with RegEx

IE11 and Edge do not implemment this property getter correctly and will return values guessed on the used protocol e.G. https://www.wheregroup.com will return 443 instead of an empty string. This behavior can result in wrongly generated Urls.

IE implementation of the pathname-getter returns the pathname without leading /. Because of this we will add the leading / when it is not present. Also we fixed accessing the regEx return value acces if nothing has matched.

Fixes #1187


werrolf:
Wrong base branch.
Not safe for `user:pass@host` and query params containing colons.

werrolf:
See [PR#1191](https://github.com/mapbender/mapbender/pull/1191), which cherry-picks the two relevant commits onto the right base branch, fixes port matching false positives and adds some reference links and other little bits.

PR #1197, opened 2019-07-11
Change visui-build.js to individual source files.

LazerTiberius:
This will make no functional difference but it will  make it a lot easier to work with this lib


werrolf:
Would prefer a glob / directory scanning thing. Unfortunately there's still files in there that we definitely do not want.

werrolf:
This will also break dist and build commands, because they pretty much empty vendor/component.  
This has only recently been fixed in [Starter commit 73134ac8](https://github.com/mapbender/mapbender-starter/commit/73134ac83eaa9555eec498e1f0f9ebbbc71908ea), which is not yet part of a tagged release.  
This will also needs extra release documentation to make people aware that they need to merge up their starters, or otherwise integrate the change.

werrolf:
Resolved more robustly in 8290d4df6722ca77e68dda3bbab44953a2c0a141
Asset requirement on `...-built.js` is expanded to list of individual known files, using vendor path instead of web/components.
Missing individual files from this expansion [are ignored](https://github.com/mapbender/mapbender/commit/7e614d7b6f03fc7b65eea4aacc5f917fe9ab6350)

Issue #1156, opened 2019-04-17
Update logo 


CharlyUnicorn:
**old logo**

The old logo is used in the mapbender-corebundle. 

**solution**

* Update image here: https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Resources/public/image/logo_mb3.png
* change name to logo_mb.png and reference it correct in the config-files (logo-path in parameters.yaml)

you can see the old logo in the administration and applications : 
* https://demo.mapbender.org/
* https://demo.mapbender.org/application/mapbender_user_yml

![mapbender_logo_website](https://user-images.githubusercontent.com/13517141/56295318-e29e6f00-612c-11e9-882d-f48f06acef3d.png)

PR #1202, opened 2019-08-07
Fix/mapbender logo

eschuettler:
The old logo is used in the mapbender-corebundle.

Fixes https://github.com/mapbender/mapbender/issues/1156


werrolf:
The logo name in ProvideBrandingPass is for recognition of "historic" logo names which get replaced with the default logo. This was done to support older configurations from a time when it was still mandatory to configure the logo name in the starter (as fom: server_logo).

We now have a default for that inside Mapbender itself.

Instead of updating the logo name in ProvideBrandingPass, I added the new name as another alternative, in addition to the old `logo_mb3.png` name.

PR #1204, opened 2019-08-29
Add scale to resolution to mapbender model ol4

frwg:
Digitizer needs this to keep min/max-scale in configuration.


werrolf:
The json.twig thing was cherry-picked onto a different branch, see #1205.

werrolf:
If we really need these methods, they should be available for both OL2 and OL4. I added / moved the implementation accordingly.

Scale is a bad metric because DPI is a bad metric used by people stuck in the paper age. My initial hope was to get rid of scales for all actual behaviour and logic, and only keep it around for informative display purposes. Time doesn't seem right yet.

PR #1205, opened 2019-08-29
Add Console Command for Creating Translation json.twig Files

werrolf:
Single commit extracted from [Pull #1204](https://github.com/mapbender/mapbender/pull/1204) to preserve reverted work.

Personally don't see the appeal.

@frwg


werrolf:
Superseded by [#1208](https://github.com/mapbender/mapbender/pull/1208). Just use a (list of) wildcard prefix(es). No need to write twigs at all anymore.

PR #1208, opened 2019-09-24
Support direct key (+wildcards) translation requirements

werrolf:
Requirements for JavaScript-available translations previously only support `.json.twig` inputs, which require maintenance and can be error prone (translation key mapping mismatches).

This pull adds optional support for Element::getAssets and Template::getAssets to include in their return value list for `trans` assets:
1) direct translation keys, e.g. `mb.core.featureinfo.error.nolayer`
2) translation key prefixes ending in a `*` wildcard, e.g. `mb.core.featureinfo.*` or even `mb.core.f*`

Wildcard prefixes expand to a list of every available translation where the key starts with that prefix. This allows extending available messages for Elements simply by adding more messages with the appropriate prefix to a catalog file (messages.en.yml et al in any activated bundle).

Note: direct key translation will throw if the translation result is equal to the input. This forces detection of untranslatable messages.

### DIFF
--- a/src/Mapbender/CoreBundle/Asset/TranslationCompiler.php
+++ b/src/Mapbender/CoreBundle/Asset/TranslationCompiler.php
@@ -5,6 +5,8 @@ namespace Mapbender\CoreBundle\Asset;
 
 
 use Symfony\Component\Templating\EngineInterface;
+use Symfony\Component\Translation\TranslatorBagInterface;
+use Symfony\Component\Translation\TranslatorInterface;
 
 /**
  * Compiles application translations for frontend consumption.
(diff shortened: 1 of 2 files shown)


src/Mapbender/CoreBundle/Asset/TranslationCompiler.php
```diff
@@ -48,4 +61,44 @@ protected function getTemplate()
...
+    protected function translatePattern($input)
+    {
+        $values = array();
+        $allMessages = null;
+        if (preg_match('/\*$/', $input)) {
+            $wildcardPrefix = rtrim($input, '*');
+            if (!$wildcardPrefix || false !== strpos($wildcardPrefix, '*')) {
+                throw new \RuntimeException("Invalid translation key input " . print_r($input, true));
+            }
+            if (!$allMessages) {
```
LazerTiberius: Am I blind or is this check unnecessary? When does $allMessage get any other value then null?

werrolf: You're right. This was supposed to be a one-time preinitialization but it didn't work that way.
Fixed in [b4ea9f4](https://github.com/mapbender/mapbender/commit/b4ea9f417612cbb5eb4a7dfa26515e89c15109b4).

PR #1215, opened 2019-11-05
Fix selection frame resets on input rotation change

LazerTiberius:
Fix erroneous form submit on enter.

Every keyup in the rotation input reset the selection frame to the center of the screen because the event argument of the event listener callback function was wrongly passed to the _updatedGeometry function
and misinterpreted as the truthy reset parameter. Now the function is wrapped and called with the correct false value for reset.
Pressing enter in one of the input fields of the print form submitted the form without the necessary logic which is bound to the print-button. Now we catch all keypress-events and check for enter(Code 13) to prevent this and call t
he correct logic.

### DIFF
--- a/src/Mapbender/PrintBundle/Resources/public/element/printclient.js
+++ b/src/Mapbender/PrintBundle/Resources/public/element/printclient.js
@@ -40,9 +40,18 @@
             $('select[name="scale_select"]', this.$form)
                 .on('change', $.proxy(this._updateGeometry, this));
             $('input[name="rotation"]', this.$form)
-                .on('keyup', $.proxy(this._updateGeometry, this));
+                .on("keyup", function () {
+                    this._updateGeometry(false);
+                }.bind(this));
             $('select[name="template"]', this.$form)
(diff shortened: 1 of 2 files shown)


werrolf:
Reimplemented in several individual commits.
1) [fix missing data if form is submitted by pressing Enter key](https://github.com/mapbender/mapbender/commit/a198d8353a82ad626e4d69c14de5f9e1405596b2)
2) [Fix printclient selection rectangle recentering on change of scale dropdown or rotation field](https://github.com/mapbender/mapbender/commit/a198d8353a82ad626e4d69c14de5f9e1405596b2)

Now that the undesired print region rectangle resets are gone, I noticed that a previously good-for-the-wrong-reasons behaviour has been lost: the print region _never_ resets. With just a little bit of zooming and panning it's very easy to "lose" the selection rectangle. If it's not visibly on screen, you have no way to find it, and you no longer have (even an accidental) way of discarding it for a new one, centered and thus visible.

I should very much like to get this resolved as well xD

Issue #1214, opened 2019-10-30
[SearchRouter] reset does not reset the selectbox properly


astroidex:
**Describe the bug**
If you have a SearchRouter configuration with Selectbox search works fine. But when you klick reset the selected option is still listed and the next search will assume that no value is selected.

**To Reproduce**
1. Create a searchRouter configuration with a selectbox

![Bildschirmfoto von 2019-10-30 15-04-18](https://user-images.githubusercontent.com/996298/67865154-208e3e00-fb27-11e9-8b24-e7fb04a28d88.png)

```
class: Mapbender\CoreBundle\Component\SQLSearchEngine
class_options:
  connection: search_db
  relation: brd.qry_gn250_p_ortslage
  attributes:
    - gid
    - name
    - gemeinde
    - hoehe_ger
  geometry_attribute: geom
form:
  name:
    type: text
    options:
      required: false
      label: Name
      attr:
        data-autocomplete: 'on'
    compare: ilike
  gemeinde:
    type: choice
    options:
      empty_value: 'Please select...'  
      choices: 
        'Bonn': Bonn
        'Bornheim': Bornheim
        'Köln': Köln
        'Siegburg': 'Siegburg'
      required: false     
results:
  view: table
  count: true
  headers:
    name: Name
    gemeinde: Gemeinde
    bundesland: Bundesland
    ewz_ger: Einwohner
    hoehe_ger: Höhe
  callback:
    event: click
    options:
      buffer: 1000
      minScale: null
      maxScale: null
  styleMap:
    default:
      strokeColor: '#00ff00'
      strokeOpacity: 1
      fillOpacity: 0
    select:
      strokeColor: '#ff00ff'
      fillColor: '#ff00ff'
      fillOpacity: 0.8
```

2. Search with selected option from selectbox -> see results
3. Reset -> see selectbox option is still selected
4. Search again -> all results are given back not only the filtered results

![Bildschirmfoto von 2019-10-30 15-05-25](https://user-images.githubusercontent.com/996298/67865166-26841f00-fb27-11e9-8903-fee02d21c898.png)

**Expected behavior**
Either 
* keep selected options after reset (only clear resultlist and inputfields)
* or clear the selectbox so the user notivces that the next search will not use the filter

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.0.8.4

**Screenshots / Logmessage**

**Data for SearchExample**

[...]

PR #1216, opened 2019-11-11
#1214 Reset does not reset the selectbox properly

pfavero-wheregroup:


werrolf:
Would much, much prefer to not have this fixed individually for one field in one Element frontend, but globally for all instances.
The "traditional" form theme is on its way out anyway as it causes nothing but trouble and hard-to-resolve conflicts. Bootstrap form theme works better, and worked better before 3.0.0.0 even came out.

werrolf:
b53e0dca97d327b337956f0fc5b1ed623387a373

PR #1219, opened 2019-11-14
Enh/sass compiler update

werrolf:
Replaces sassc compiler dependency with [wheregroup/assetic-filter-sassc](https://packagist.org/packages/wheregroup/assetic-filter-sassc) plus [wheregroup/sassc-binaries](https://packagist.org/packages/wheregroup/sassc-binaries).

Allows using native sassc, e.g. available on Debian / Ubuntu via apt get. Set the path to your desired binary via parameter `mapbender.asset.sassc_binary_path`, usually to `/usr/bin/sassc`. This disables autopicking of the binary from one of the bundled ones.

The compiler change fixes the long broken minified CSS output in `prod` environment, aka `app.php`.

On Linux, outside of prod, results are identical (as per md5sum).  
This change needs verification on Windows and MacOS.  
The compiler is also used to validate the Application CSS form field in the backend. This should be reverified with / without deliberate errors in the CSS input.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,8 @@
         "wheregroup/codemirror": "5.x",
         "components/underscore": "1.x",
         "wheregroup/open-sans": "1.x",
+        "wheregroup/assetic-filter-sassc": "^0.0.1",
+        "wheregroup/sassc-binaries": "^0.0.1",
 
         "mapbender/fom": "~3.1.2 || ^3.2.1",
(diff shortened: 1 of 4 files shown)


werrolf:
Testing note: because this changes dependencies, I suggest a
```sh
bin/composer require 'mapbender/mapbender:dev-enh/sass-compiler-update@dev as 3.0.8.99'
```
in the root package. This should be the easiest way to switch to this version.

PR #1235, opened 2020-04-16
Fix/reset search router

pfavero-wheregroup:
#1214 Reset does not reset the selectbox properly


werrolf:
This is fixed in 3.0.8.6-RC2 globally, not just for SearchRouter, without escalating form theme dependencies. See 741fde8ad0377a3491d6d82a440b5c9770b5a8ce

Issue #1236, opened 2020-04-24
Saving of WMS Instance with grouped Layers scrambles configuration


frwg:
**Describe the bug**
A freshly added WMS layerset instance with grouped layers does change its configuration of it's `fold/allow` flags. Checked layer groups lose their flags in favor of the respective layers they consist of (but not consistently)

**To Reproduce**
1. Add `https://geodaten.metropoleruhr.de/alkis/alkis?Service=WMS&Version=1.3.0&REQUEST=getCapabilities` to sources
2.  Add instance to layerset of an application
3. click "save"

**Expected behavior**
Visible configuration stays exactly as it was before pressing the save button 

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.0.8.5 
 - Bug does not appear in Mapbender 3.0.8.4 and Mapbender 3.1

**Screenshots / Logmessage**
before saving:
![vorher](https://user-images.githubusercontent.com/44997160/80211805-a7375a00-8636-11ea-98d4-9a8bc0d1ac32.png)
after saving:
![nachher](https://user-images.githubusercontent.com/44997160/80211812-a999b400-8636-11ea-8811-3386e4f59d67.png)

frwg:
It looks like there is a simple way to prevent this: After creating the instance, don't directly save but leave the menu (cancel) and re-open the layerset. This  should make for a safe save.

frwg:
Bug detected: Within the WMS instance config, the application expects the WMS-Layers of a WMS Source to be ordered according to their ids when saved. The created form relies on this and the blank form's instance table is ordered like that. Iinternally, the inputs of the form are annotated with a counte:, the first row's inputs with 0, the second with 1, the third with 2 and so on. However, after creating an instance source, the configuration menu's instance table's inputs counters are messed up. They do not ascent starting with 0, but rather descend starting with the highest number. Saving this leads to distorted configuration, with the first row having the values of what was the last row before and so on.  Responsible for this is the incomplete ordering of the wmsinstances layers - the query orders tham after priority only, not by id, as the application expects it.

werrolf:
Issue is missing initialization of the instance layer order (which is user-editable). This initialization is currently only performed client-side in the form JavaScript, which is bad practice.

astroidex:
tested with 3.0.8.6rc1 works fine

PR #1239, opened 2020-04-28
adjust ORM annotation of WMS InstanceLayer join to guarantee the Laye…

frwg:
…rs to be ordered by id

Within the WMS instance config, the application expects the WMS-Layers of a WMS Source to be ordered according to their ids when saved. The created form relies on this and the blank form's instance table is ordered like that. Internally, the inputs of the form are annotated with a counter:, the first row's inputs with 0, the second with 1, the third with 2 and so on. 

However, after creating an instance source, the configuration menu's instance table's inputs counters are messed up. They do not ascent starting with 0, but rather descend starting with the highest number. Saving this leads to distorted configuration, with the first row having the values of what was the last row before and so on.  Responsible for this is the incomplete ordering of the wmsinstances layers - the query orders them after priority only, not by id, as the application expects it.

see https://github.com/mapbender/mapbender/issues/1236


werrolf:
Merged into release/3.0.7. Github no longer allows a base change because there are no more new commits.

PR #1252, opened 2020-08-14
Fix for wrong icon scale and not printing circle features

LazerTiberius:
* Fix not printing circles 
* Fix not respecting icon scale in print


werrolf:
Yes.

Issue #1263, opened 2020-11-03
Mapbender 3.2.2: app/console mapbender:config:check says PHP 5.5.4


astroidex:
**Is your feature request related to a problem? Please describe.**
Mapbender 3.2 needs PHP >= 7.1 (see release mail https://lists.osgeo.org/pipermail/mapbender_users/2020-July/005813.html)

But when you check with 3.2 it says Required Version PHP 5.5.4

**Describe the solution you'd like**
Please update the config check to the Required Version PHP

see https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Command/ConfigCheckCommand.php#L100

**Versions **
 - Mapbender: 3.2.2

astroidex:
Created PR https://github.com/mapbender/mapbender/pull/1266

werrolf:
If you want to know the php version, type `php --version`.

PR #1266, opened 2020-11-04
update app/console mapbender:config:check >= 7.1.0

astroidex:
* Mapbender 3.2 needs PHP >= 7.1.0 
* app/console mapbender:config:check still checked for >= 5.4.4
* this was fixed. Checked with composer.json to check for the right PHP version
* see https://github.com/mapbender/mapbender/issues/1263


rschwammborn (approved):

werrolf:
We will not do this.
Instead, we will remove the legacy crutch hack of defining a "platform" in composer.json, tricking composer into thinking the PHP version is something it isn't. We require PHP 7.1. You will not be able to install Mapbender 3.2+ on anything less, with the platform override removed.

If you can't install, checking after install if you could have installed becomes redundant.

werrolf:
Fun fact: if you have less than PHP7.1, you can't even run this command, because numerous container initializers will fail.

PR #1270, opened 2020-11-10
Feature/feature info with highlight layer

frwg:
Adds capability to extract, render and highlight individual geometric features from a HTML FeatureInfo response.
FeatureInfo HTML must contain elements with `data-geometry` attributes, which must contain valid WKT or EWKT geometries. The same Element may also have `data-srid` attributes, which should name the CRS of the WKT geometry. This property is ignored for if `data-geometry` contains an EWKT.

If `data-geometry` only contains a WKT, and the element does not also have a `data-srid` attribute, geometries are assumed to be in the current CRS of the map view.

The entire highlighting functionality is controlled by a new boolean `highlighting` option in the FeatureInfo element configuration (yaml + backend editing capabilities).

Highlighting only works when using current map engine. We do not provide an implementation for Openlayers 2.


werrolf:
Adds capability to extract, render and highlight individual geometric features from a HTML FeatureInfo response.
FeatureInfo HTML must contain elements with `data-geometry` attributes, which must contain valid WKT or EWKT geometries. The same Element may also have `data-srid` attributes, which should name the CRS of the WKT geometry. This property is ignored for if `data-geometry` contains an EWKT.

If `data-geometry` only contains a WKT, and the element does not also have a `data-srid` attribute, geometries are assumed to be in the current CRS of the map view.

The entire highlighting functionality is controlled by a new boolean `highlighting` option in the FeatureInfo element configuration (yaml + backend editing capabilities).

Highlighting only works when using current map engine. We do not provide an implementation for Openlayers 2.

PR #1272, opened 2020-11-13
enhance style of scaleselector to prevent linebreak between label and…

frwg:
… div


frwg:
already solved in master

PR #1279, opened 2020-11-24
Rename "Redlining" element to "Sketch"

werrolf:
We've internally decided to distance ourselves from problematic "Redlining" wording.

This pull renames the class and all related assets to "sketch".

There's a new [class migration entry](https://github.com/mapbender/mapbender/commit/b35dac66b9d5b37190dea8511312052dd217d2be) that allows existing applications to work.

Custom child classes of the PHP Redlining class should update their base classes.
Custom extensions of the "mbRedlining" javascript widget will be broken by this change and will need to change their parent widget to "mbSketch".


DeEgge:
👏

Issue #1280, opened 2020-11-25
[Frontend][PrintClient] Hochwert in print output is switched 


astroidex:
**Describe the bug**
In Version 3.0.8.6 the hochwert in the print output is not correct - They are switched. The value at the upper right should be at the lower left.

This is fixed in 3.2 see:
https://github.com/mapbender/mapbender/commit/2a01f7297b49c022878737d815ba5497f683eb3c#diff-03a4c9b662d234d98b9539f6ec141786e7c96fe8efb1e80abbdc0dded43abd6e

**To Reproduce**
1. Go to https://version-lts.mapbender.org/v3.0.8.6/application/mapbender_user_yml_printClient
2. Chosse print a4_landscape_offical coordinates
3. create a pdf and see the error

**Expected behavior**
Correct labeling of the print output.

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.0.8.6 (error does not occur in 3.2)

**Screenshots / Logmessage**
frame with coordinates:
- lower left: 362590.89 5620040.05
- upper right: 369625.93 5623932.07
![frontend_print](https://user-images.githubusercontent.com/996298/100194571-3884ec80-2ef6-11eb-8875-37e11ae49043.png)

See error
![print_pdf_with_flipped_coordinates](https://user-images.githubusercontent.com/996298/100194605-42a6eb00-2ef6-11eb-8944-398c59f29cec.png)

rschwammborn:
Hello @astroidex  ,
thanks for reporting the bug - I created a ticket in the internal Mapbender ticket system.
We'll keep you updated!

PR #1283, opened 2020-12-01
change values of fieldDataMapping according to min/max values of

frwg:
Fix to this ticket: https://github.com/mapbender/mapbender/issues/1280

### DIFF
--- a/src/Mapbender/PrintBundle/Component/PrintService.php
+++ b/src/Mapbender/PrintBundle/Component/PrintService.php
@@ -612,29 +612,33 @@ class PrintService extends ImageExportService implements PrintServiceInterface
         }
 
         $efData = $jobData['extent_feature'];
+        $minX = min(array_map(function($coordinate){ return $coordinate["x"]; },$efData));
+        $minY = min(array_map(function($coordinate){ return $coordinate["y"]; },$efData));
+        $maxX = max(array_map(function($coordinate){ return $coordinate["x"]; },$efData));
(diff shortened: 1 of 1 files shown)


astroidex:
I tested the PR. It works fine.

src/Mapbender/PrintBundle/Component/PrintService.php
```diff
@@ -612,29 +612,33 @@ protected function addCoordinates($pdf, $template, $jobData)
         }
 
         $efData = $jobData['extent_feature'];
+        $minX = min(array_map(function($coordinate){ return $coordinate["x"]; },$efData));
+        $minY = min(array_map(function($coordinate){ return $coordinate["y"]; },$efData));
+        $maxX = max(array_map(function($coordinate){ return $coordinate["x"]; },$efData));
+        $maxY = max(array_map(function($coordinate){ return $coordinate["y"]; },$efData));
         $fieldDataMapping = array(
```
werrolf, line 619: Min / max of the coordinates _do_ _not_ produce the bottom left / top right corner of the printed map area. Especially not for a rotated print.

werrolf:
Fixed in ee257e1a144bf2d61ade8a2d45181ab8a224b21a

Openlayers 2 bounds (which we use for print selection) order their coordinates top left => top right etc
Openlayers 4 bounds use bottom left => top left etc

Fix does not min / max (which would produce more errors if print uses rotation), but adjusts extracted coordinates to start bottom left on Openlayers 2.

PR #1344, opened 2021-04-15
 sketch translation suggestion: stop_drawing: Zeichnen unterbrechen

astroidex:
stop_drawing: Zeichnen unterbrechen instead of Zeichnen stoppen, as you only pause the drawing and can continue


astroidex:
changed to Zeichnen abbrechen - suggestion by @rschwammborn

rschwammborn (approved):

PR #1345, opened 2021-04-15
Fix config request content type

swinkelmann:
'json' is not a valid content type. 
This causes problems whenever a firewall with a strict set of rules is involved.


werrolf (approved):
Agreed.
Actually, we can later set this to false, because `contentType`, even though it is always supplied as a request header and may then cause problems, is only relevant for data sent to the server -- we are not sending any.

PR #1360, opened 2021-05-28
Login backdrop image customizability

werrolf:
The redesigned backend adds a large backdrop image to the login area.

Changes in this pull make this image customizable on the project level.

A new container parameter `branding.login_backdrop` is introduced. This should name an image asset located inside the web directory (=docroot), and should be a relative url path.

The image covers half the screen horizontally, is scaled to full height and centered. Thus the ideal aspect ratio is half the anticipated screen aspect ratio (8:9 for typical 16:9 FullHD screens).

### DIFF
--- a/src/Mapbender/CoreBundle/Extension/BrandingExtension.php
+++ b/src/Mapbender/CoreBundle/Extension/BrandingExtension.php
@@ -1,0 +1,47 @@
+<?php
+
+
+namespace Mapbender\CoreBundle\Extension;
+
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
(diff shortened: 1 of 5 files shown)


werrolf:
### Example usage
File placed into web directory:
```sh
application$ ls -l web/*.jpg
-rw-rw-r-- 1 ron ron  27672 Mai 28 09:39 web/pirate-cat.jpg
```
File referenced in app/config/parameters.yml
```yaml
parameters:
    <...>
       branding.login_backdrop: pirate-cat.jpg
```
Result:
Screenshot_2021-05-28_09-42-44

Issue #1369, opened 2021-06-30
[backend] missing colorpicker assets / broken FeatureInfo backend form if installed in a sub-path URL


astroidex:
**Describe the bug**
If you are in the backend and want to edit / add a featureInfo element. You get an error message $pickers..colorpicker is not a function and the element can not be added

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.2.5

astroidex:
Fixed with the following commit https://github.com/mapbender/mapbender/commit/f66f648970d924884afc3b32c231b57f8f769687

astroidex:
tested with 3.2.6: works fine

PR #1371, opened 2021-07-06
Fix missing colorpicker assets on windows in frontend

astroidex:
- css can not be found on windows in frontend
- was fixed in manager.html.twig 
- see ticket https://github.com/mapbender/mapbender/issues/1369


astroidex:
is fixed

werrolf:
Not fixed.
NOTE: this is not about Windows. This happens on any OS, if the project is served from a sub-url (`http://hostname/something/app.php`).

PR #1422, opened 2022-07-18
Sketch: multiple, configurable drawing colors + frontend customizable color

werrolf:
Adds a multi-entry color pallette to Sketch element to replace the previous hardcoded red default.

Screenshot_2022-07-18_12-47-11

Main pallette entries are predetermined by element configuration. The default (as seen above) is to offer red, green and blue.
Additionally, users can edit their own extra color with a color picker.

Clicking on one of the color buttons determines the color for the next feature drawn, or, if currently editing a feature (click on edit icon in table listing), immediately updates that feature to that color.

If the element is configured with only a single color _and_ the frontend color picker is disabled, the pallette region is not shown.

For yaml elements:
```yaml
    class: Mapbender\CoreBundle\Element\Sketch
    colors:          # new; optional list of strings; use CSS hex notation; omit completely for defaults
      - '#4488ff'
      - '#ff8844'
    allow_custom_color: true           # new; default true; set to false to disable frontend color picker
```

_Note_ that in the backend form, `colors` is a scalar input field. Color values must be comma separated.

### DIFF
--- a/src/Mapbender/CoreBundle/Element/Sketch.php
+++ b/src/Mapbender/CoreBundle/Element/Sketch.php
@@ -4,9 +4,11 @@ namespace Mapbender\CoreBundle\Element;
 
 use Mapbender\Component\Element\AbstractElementService;
 use Mapbender\Component\Element\TemplateView;
+use Mapbender\CoreBundle\Component\ElementBase\ConfigMigrationInterface;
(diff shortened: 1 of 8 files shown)


astroidex:
Please check draw point, The Point is always black althought you picked a color

PR #1425, opened 2022-08-12
Activation of the supplied Dutch translation

ThorstenHi:
A [Dutch translation](https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Resources/translations/messages.nl.yml) has been part of Mapbender for a long time. With this fix it can be used again.


werrolf:
If you don't like Mapbender respecting your browser language, you can turn it off via configuration.
```yml
parameters:
    <...>
    mapbender.automatic_locale: false
```

werrolf (approved):

Issue #1276, opened 2020-11-20
[frontend][SearchRouter] mandatory field are not focussed if there are not filled


astroidex:
**Describe the bug**
In searchrouter you can define mandatory fields. When you use a mandatory field the search will not run before you fill the field. Mandatory fields are makred with a red start. 
Problem: when you click ok, you do not see that the search does not start and is still waiting for the user to fill the field

For testing the following yml can be used: https://github.com/mapbender/mapbender-workshop/blob/release/3.2/app/config/applications/mapbender_demo_nrw.yml

**To Reproduce**
1. Go to SearchRouter
2. Go to a search with mandatory fields
3. Klick search without filling the fields
4. See error: nothing is happening - no highlight of the filed/choice and no info that input is needed

**Expected behavior**
Mandatory field should be somehow highlighted to show the user that an input is expected

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.2.2

**Screenshots / Logmessage**

**Additional context**

thzanger:
Hello @astroidex, thank you for this information. The issue has been forwarded into the internal Mapbender ticket system.

werrolf:
Missing red asterisks for required fields have been fixed in Mapbender v3.2.3.

PR #1444, opened 2023-03-08
Fix missing visual feedback when submitting invalid form

Phocacius:
in Search Router and WMSLoader

refs #1276

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/views/Element/search_router.html.twig
+++ b/src/Mapbender/CoreBundle/Resources/views/Element/search_router.html.twig
@@ -8,7 +8,7 @@
         {% endfor %}
     </div>
     <div class="search-action-buttons">
-        <span data-action="reset" class="btn btn-sm btn-default">{{ 'mb.actions.reset' | trans }}</span>
-        <span data-action="search" class="btn btn-sm btn-default">{{ 'mb.actions.search' | trans }}</span>
+        <button data-action="reset" class="btn btn-sm btn-default">{{ 'mb.actions.reset' | trans }}</button>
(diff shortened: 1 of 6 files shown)


astroidex (approved):
Tested the PR. it works as exprected. Focus and message on mandatory fields if no input is provided or no selection was made.

src/Mapbender/CoreBundle/Resources/views/Element/search_router.html.twig
```diff
@@ -8,7 +8,7 @@
         {% endfor %}
     </div>
     <div class="search-action-buttons">
-        <span data-action="reset" class="btn btn-sm btn-default">{{ 'mb.actions.reset' | trans }}</span>
-        <span data-action="search" class="btn btn-sm btn-default">{{ 'mb.actions.search' | trans }}</span>
+        <button data-action="reset" class="btn btn-sm btn-default">{{ 'mb.actions.reset' | trans }}</button>
```
werrolf, line 11: Does this work as expected?
HTML button with no `type` attribute defaults to type="submit" and will implicitly submit the enclosing form. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button).

Can you please verify that
* Clicking on the search button only executes the search once
* Clicking on the reset button does not execute a search

astroidex: hi @werrolf,
I checked out the branch and followed your instruction (with F12 network opened). Tested with Firefox

-     Clicking on the search button only executes the search once -> yes
-     Clicking on the reset button does not execute a search -> yes

So it looks it works fine

PR #1447, opened 2023-03-30
Add option to disable newly added layers during WMS update in GUI and CLI

Phocacius:


PhilVanB:
Looks good to me. I like the use of "new" PHP syntax in particular: 

https://github.com/mapbender/mapbender/pull/1447/commits/549b5edeb18ec6323cb1bed1304c3698f8fc9e2c#diff-5111458587caae137013ba11fd1721a7c6f58c7c788168769937bb98027438a5R132

I would not place an if-statement in the parameter transfer though:

https://github.com/mapbender/mapbender/pull/1447/commits/549b5edeb18ec6323cb1bed1304c3698f8fc9e2c#diff-5111458587caae137013ba11fd1721a7c6f58c7c788168769937bb98027438a5R115

I'd put them in an extra line before. We would then have a clearer arrangement and the code line is not that long: -> so _git diff_ cannot raise ugly line breaks that easy.

If I'd had to update the changelog, my commit message would be "Update changelog" instead of "Extend changelog":

https://github.com/mapbender/mapbender/pull/1447/commits/8d7256728b10dcebd4ad7f7039e56dae5d28ca21

Phocacius:
@PhilVanB I agree with the splitting of the too long line, I changed it. For the Changelog commit message: I'll remember for next time.

Also, after testing by @astroidex the option was split into two options for deactivation and deselection, if you want to can have another look, it works the same as before though.

PhilVanB:
:smile: Code is all fine now!

astroidex (approved):

Issue #1441, opened 2022-12-02
[Backend] WMS can not be loaded/updated


astroidex:
**Describe the bug**
When you try to load a WMS or update a WMS in Mpabender 3.3.2  you get an error message (with database PostgreSQL).

![Bildschirmfoto vom 2022-12-02 11-08-02](https://user-images.githubusercontent.com/996298/205268618-bf53ddcd-19b4-4f73-a81e-cc636bf5f197.png)

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.3.2
 - Database: PostgreSQL

astroidex:
The error was fixed see PR https://github.com/mapbender/mapbender/commit/3aa6694ebba3c437753c4b0d6ae9ba86e7b01bfc

you can replace the file with the following
https://github.com/mapbender/mapbender/blob/3aa6694ebba3c437753c4b0d6ae9ba86e7b01bfc/src/Mapbender/WmsBundle/Entity/WmsLayerSource.php

The fix will be available in the next version 3.3.3.

rschwammborn:
Tested this (just in case): Can confirm that it has been fixed in 3.3.3. I'm closing the issue.

PR #1448, opened 2023-04-04
Type check boolean and int values when importing WMS into database

Phocacius:
refs #1441


astroidex:
@Phocacius I tested the PR with the Services that makes trouble. Works fine. I also tested some other WMS. they all work fine and can be added.

astroidex (approved):

PR #1453, opened 2023-04-06
Improve scale select and scale display elements

Phocacius:
- Add thousand separators in all places within the two elements (using JS's [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) and PHP's [NumberFormatter](https://www.php.net/manual/de/numberformatter.create.php))
- Fix bug where the current scale in the scale select element became blank after selecting a scale
- Make the scale prefix localisable in scale display
- Set default prefix in scale display to a localised string
- Localised backend form of scale display and added help texts


astroidex:
- delimiter should be . not ,
Bildschirmfoto vom 2023-04-06 14-29-20

astroidex (approved):

Phocacius:
@astroidex That depends on your language settings. If your language is set to English, then , is the correct grouping separator. if your language is set to German, the . is used as grouping separator

astroidex:
I see

in my case i use 
- firefox / english -> shows ,
- chromium german -> shows .

, makes no sende at all. What can we do?

Phocacius:
That makes perfect sense since this is how those languages work :)

See e.g. https://language-boutique.com/lost-in-translation-full-reader/writing-numbers-points-or-commas.html

PR #1454, opened 2023-04-06
[SearchRouter] Zoom to feature automatically if there is only one result

### LINKED ISSUE #1411: SearchRouter zoom to feature directly if there is only one result
**Is your feature request related to a problem? Please describe.**
SearchRouter lists all the results in a table. On click on a result Mapbender zooms to the object. 

It would be nice to zoom directly to the result item if there is only one result.

**Describe the solution you'd like**
It would be nice to zoom directly to the result item if there is only one result.

Phocacius:
closes #1411


astroidex (approved):

astroidex:
works fine

PR #1455, opened 2023-04-11
Security settings could not be saved if a user or group where access control has been previously defined is deleted

Phocacius:
The (deprecated) symfony/acl-bundle has a bug that prevents an access control list (ACL) from being saved,
when a security identify (e.g. a user) was deleted that had an entry in the component's ACL and was not the
last one in the list. The ACL expects the ace_order column to start with zero and increase one by one,
otherwise array indexing fails.

For example: An application has access rights for user A, B and C. User B is deleted. Then, the acl_entries
table will have only entries for users A (ace_order 0) and C (ace_order 2) which can't be saved anymore.

Implemented workaround: After deleting groups or users (and after manual trigger), a service goes through the whole acl_entries table and resets the indices of entries where the ace_order contains gaps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## dev-master @ df7481c88
+## v3.3.4-RC1
 * Add Ukranian translations by sacredkesha ([PR#1442](https://github.com/mapbender/mapbender/pull/1442))
 * Enable Ukranian in locale auto-detection ([#1443](https://github.com/mapbender/mapbender/issues/1443))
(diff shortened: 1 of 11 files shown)


astroidex:
Hi @Phocacius , i tested the PR

There is still strange behaviour. What I did.

1. ran the command app/console mapbender:security:fixacl
2. added User/group to an application
3. deleted the group
4. user & group are not listed at the application
5. when I try to add more users to the application, I get an error on save (see screenshot)
6. after refresh the users are not listed for the application

Bildschirmfoto vom 2023-04-12 10-35-44

astroidex (approved):

PR #1456, opened 2023-04-11
[Sketch] Remove buggy geometry type 'text', all its features are already represented by 'point'

Phocacius:


astroidex:
I like the type text and qould keep it but only do not show a circle at the position (only show the cirlce when you move the text)

Phocacius:
That would have the problem that people wouldn't know where the text should point to, although in case of larger polygons that might not be a problem.
I'd still vote in favour of removing the type, it's confusing that point and text do almost the same thing. I'd rather have an additional transparency slider in the color selection so you'd could set the opacity to 0 in a point geometry type to achieve a hidden marker.

astroidex (approved):

Issue #1404, opened 2022-01-19
[mobile application] theme dialog should not close after selecting a theme


astroidex:
**Describe the bug**
[mobile application] theme dialog should not close after selecting a theme. It should be possible to select several themes at a time.

**To Reproduce**
1. Go to https://demo.mapbender.org/application/mapbender_mobile_yml or https://demo.mapbender.org/application/mapbender_mobile_yml6
2. Click on 'Themes' and disable/enable a theme
4. See error: Dialog closes after selecting a theme
5. This behaviour is rightt vor the Basesourceswitcher but in the theme section you would like to select more than one theme at a time (this was the behavior in older Mapbender version 3.0.5 https://wesseling.wheregroup.com/mapbender3/application/wesseling_mobile) 

**Expected behavior**
When you are in the theme section it should be possible tho un/check themes without closing the dialog

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.2.8

astroidex:
A solution could be in file 

- https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js#L385
- https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js#L401

Remove the following lines

            //if (this._mobilePane) {
            //    $('#mobilePaneClose', this._mobilePane).click();
            //}

PR #1457, opened 2023-04-12
Fixes in LayerTree element concering sorting

Phocacius:
- Add jquery-ui-touch-punch library for touch support
- Do not close mobile pane when selecting a theme (see #1404)
- Do not disable sortable functionality in layertree when on devices that support touch

- Fix z-index of mobile pane so it is always above map elements like the simple search


astroidex:
- **workss fine** Add jquery-ui-touch-punch library for touch support
- **workss fine** Do not close mobile pane when selecting a theme 
- **workss fine** Do not disable sortable functionality in layertree when on devices that support touch
- **workss fine** Fix z-index of mobile pane so it is always above map elements like the simple search

astroidex (approved):

PR #1463, opened 2023-04-28
Color selection in FeatureInfo and Ruler

Phocacius:
* [FeatureInfo] Redesigned highlight color selection: Opacity is now part of the color selection. Stroke and fill opacity can now be selected independently 
* [Ruler] Added configuration options:
    - help (default: 'mb.core.ruler.help')
    - fillColor (default: 'rgba(255,255,255,0.2)')
    - strokeColor (default: '#3399CC')
    - strokeWidth (default: 1)
    - strokeWidthWhileDrawing (default: 3)
    - fontColor (default: '#000000') 
    - fontSize (default: 12)
* [Ruler] Improved value display. Thousand separators are now used, the users's locale decimal separator is respected, in precision setting 'auto' the number of fractional digits displayed now varies depending on the number 
* StyleUtil: Add support for #rrggbbaa color format


astroidex:
- thanks for the changes, very good to use #rrggbbaa color format and to use thousand separators (info: we have colorpixer in digitizer too, so maybe change it there too use the same concept)

requested changes
- 1.  [Ruler]  suggestion to set the default 2 for strokeWidth -> at the moment it is default: 1

2. element interaction got lost in 3.3.4 (was there in 3.3.3)
When you use measure, print,coordinate utility the featureInfo should not fire requests, see last demo 3.3.3:
- https://demo.mapbender.org/application/mapbender_user_yml (activate FeatureInfo, use measure -> no featureInfo while you measure)

astroidex:
Looks like the Font color you can define is not the font color but the font-outline color

I would suggest to change it to font-color. Maybe in future release we could also add the font-outline color if it is needed

Example
- font color set to yellow in the elememt definition, but it is used as outline color

ruler-font-color

astroidex:
There is a featureInfoHighlight Demo in the workshop Bundle (for internal use)

- https://github.com/mapbender/mapbender-workshop/blob/master/app/config/applications/mapbender_user_featureinfo_highlight.yml
- Demo is updated for 3.3.4

Phocacius:
Text color is fixed. Merging now so we can create the RC2.

astroidex:
I tested with 3 results. Hover works fine when I only hover in the dialog. But if an area was "hovered" in the map first the hover is still there when you hover in the dialog

fi_hover

Every object needs a unique id in the div see definition:
fi_hover

PR #1470, opened 2023-05-10
Fix search router highlighting was reset after hovering over another item

Phocacius:
Steps to reproduce:
- setup search router with different highlight styles for 'select' and 'temporary'
- Click on a search result to select it
- Hover over the item list without selecting another result
- After mouseout from the previously selected result it returned to the default selection state instead of staying in select mode


astroidex:
Very good. I tested the PR. Itw works as expected.

astroidex (approved):

PR #1471, opened 2023-05-10
Fixes for FeatureInfo in Mobile Template

### LINKED ISSUE #1467: [mobileTemplate] featureInfo does not work
**Describe the bug**
When you use the mobiile demo featureInfo is active by default. Butno featureInfo information is shon.

When you add a button for featureInfo. It will open a dialog when you activate it instead of fireing a featureInfo command.

 - Mapbender: 3.3.4
 - Browser firefox, chrome

Phocacius:
- Feature Info now working again in mobile template. The problem here was, that `util:checkReponsiveVisibility` checked only for visibility, but not why the element was invisible. Added check for ''hide-screentype-*" classes.
- When feature info is managed via a button, it is now prevented that the mobile popup appears each time the state is toggled

fixes #1467


astroidex:
Tested the PR. Works fine.

astroidex (approved):

PR #1475, opened 2023-05-24
[SearchRouter] Fix sarch failed due to cached csrf tokens in production environment

Phocacius:
Regression bug after #1458: The main mapbender site is cached in production mode which includes the search router form. Therefore, for the first user, where the page was regenerated, everything was working as expected since the page was regenerated and the csrf token properly stored in the session. For the subsequent users, the response got cached, resulting in them receiving the same token as the first user which did not pass the csrf check though, since no token was stored for that user.

Implemented solution: Upon startup, an ajax request (which is never cached) queries for a current csrf token  and stores it in the form.


ThorstenHi (approved):
Works as desired.
Thank you!

astroidex (approved):

PR #1478, opened 2023-05-25
[LayerTree] Correctly show folder state (opened/closed) when thematic layers are active

Phocacius:


astroidex:
I tested the behaviour. Works fine.

astroidex (approved):

PR #1487, opened 2023-06-22
Update ExportResponse.php - enable Usage of xslx as file type

frwg:


src/FOM/CoreBundle/Component/ExportResponse.php
```diff
@@ -158,6 +179,7 @@ public function setType($type)
                 $this->headers->add(array('Content-Type' => 'text/csv;charset=' . self::UTF_16_LE));
                 break;
             case self::TYPE_XLS:
+            case self::TYPE_XLSX:
```
Phocacius, line 182: according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types the MIME type for XLSX is `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`

src/FOM/CoreBundle/Component/ExportResponse.php
```diff
@@ -110,6 +120,17 @@ public function setXls(array &$data){
...
     }
 
+     /**
+     *
+     * @param array $data
+     * @return void
+     */
+
+    public function setXlsx(array &$data){
+        $xlsx = SimpleXLSXGen::fromArray( $data );
```
Phocacius, line 130: Where does `SimpleXLSXGen` come from? It's not imported and there is no file with that name currently present

frwg: Ah, okay, then nevermind. Just assumed it was a library dependent on Mapbender.

composer.json
```diff
@@ -19,6 +19,7 @@
         "doctrine/doctrine-bundle": "^1.12 || ^2",
         "symfony/swiftmailer-bundle": "^2.3 || ^3",
         "sensio/framework-extra-bundle": "^3 || ^4 || ^5 || ^6",
+        "shuchkin/simplexlsxgen": "^1.2"
```
Phocacius, line 22: you're missing a comma here which would render the json invalid. And the current version of the library is 1.3, the API stayed the same so I'd say it's safe to just update it

Phocacius (approved):

PR #1489, opened 2023-06-22
[Map element] Handle incorrect EPSG codes

Phocacius:


astroidex:
Tested. Works fine.

Phocacius:
@astroidex Did you check out the correct branch during testing? The screenshot you show was the old, confusing behaviour.  The new message should shouw `Unsupported projection " + projOrSrsName.toString() + ", falling back to EPSG:4326. Check configuration of map element.`

Phocacius:
@astroidex The display of the multiple messages is fixed. Merging now so I can create the next release candidate

PR #1490, opened 2023-06-23
Update LayerRendererGeoJson.php

frwg:
enable rendering of multipoint geometry


src/Mapbender/PrintBundle/Component/LayerRendererGeoJson.php
```diff
@@ -296,38 +299,59 @@ protected function drawMultiLineString($canvas, $geometry)
      */
     protected function drawPoint($canvas, $geometry)
     {
+        // promote to single-item MultiLineString and delegate
```
Phocacius, line 302: the only minor point: The comment says MiltiLineString, but you're promoting to MultiPoint here

Phocacius (approved):

PR #1493, opened 2023-06-29
update WMSLayerSource instead of delete and recreate - on Source Update

frwg:
Please review the code thoroughly before mergin - ideally we would set up an automated test for this, but since this must include a database, we should discuss the issue first.


frwg:
Use https://github.com/mapbender/mapbender/pull/1497 instead

PR #1508, opened 2023-10-20
Fix search

Phocacius:
Replace deprecated QueryBuilder::execute() by executeQuery()

Error introduced after updating of doctrine DBAL


astroidex (approved):
tested. Works fine!

PR #1511, opened 2023-10-24
Fix remaining old class references

Phocacius:
fixes 
- error when adding or editing users
- error when editing security settings for an application 
- ajax error when adding to print queue


astroidex:
Testing
- add user works fine
- add group works  fine

error: Add Class ACL Application 
- /acl/edit?class=Mapbender\CoreBundle\Entity\Application
- browser freezes and does not add the user/group to 
- in the list the email is shown for user and not the name of the user

Phocacius:
@astroidex Problems should be fixed, please test again.

astroidex:
Tested again. Works fine. 
I did not test selfregister and resent password. @Phocacius  id you test that?

astroidex (approved):

Phocacius:
No, I did not test that yet.

PR #1517, opened 2023-10-30
Remove underscore dependency

Phocacius:
The library was only used sparsely and was not worth the effort of keeping up to date. The following replacements can be used:

- `_.assign`, `_.extend`: `Object.extend`
- `_.debounce`: `Mapbender.Util.debounce`
- `_.difference`: Write manully (one-liner)
- `_.each`, `_.forEach`: `Array.prototype.forEach` or `JQuery.each`
- `_.findWhere`: `Mapbender.Util.findFirst`
- `_.mapObject`: Write manully (three-liner)
- `_.object`: Write manully (three-liner)
- `_.omit`, `_.filter`: `Mapbender.Util.filter`
- `_.uniq`: `Mapbender.Util.array_unique`


PhilVanB:
Good job. Approved.

PR #1529, opened 2023-12-11
Update IconPackageFa4.php iconEarth fas fa-earth-africa instead of far

astroidex:
- must be fas fa-earth-africa


src/Mapbender/CoreBundle/Component/IconPackageFa4.php
```diff
@@ -127,7 +127,7 @@ public function getIconMarkup($iconCode)
             case 'iconMapPin':
                 $class = 'fas fa-map-pin'; break;
             case 'iconEarth':
-                $class = 'far fa-earth-africa'; break;
+                $class = 'fars fa-earth-africa'; break;
```
Phocacius, line 130: should be "fas", right?

astroidex:
yes. sorry

PR #1530, opened 2023-12-15
Fix/bootstrap bugs

PhilVanB:
Fixes issues from ERM #8028


astroidex (commented):
Hi @PhilVanB I tested the PR. It is fine. Added some more  comments to #8028

PR #1536, opened 2024-01-12
element ruler: change ruler to be able to create multiple geometries while measuring

astroidex:
Allows to draw multiple geometries. Generates calculation for every single geometry. 

Please check whether the layout is ok. Any ideas?

Ideas for future developments
- highlight geometry on mouse-over
- calculate total (length  or area of the geometries)
- allow to delete geometry from the list
- allow to modify geometries


astroidex:
After discussion we decided not to take this enhancement in the core.

PR #1537, opened 2024-01-29
Fix path to LDAP-userclass

PhilVanB:
- Fixes the path to the user class of the currently used MapbenderLDAPBundle.
- Allows to secure an application with an LDAP-User again.


astroidex:
Please add a Changlog entry with the PR

Phocacius (approved):

astroidex:
was also merged to develop

PR #1538, opened 2024-01-31
New console command `mapbender:normalize-translations`

Phocacius:
The console command `mapbender:normalize-translations` does three things:
- sort all translation keys in all bundles in the `Mapbender` and `FOM` namespace in the same order as in the original translation file (usually English)
- show keys that are present in the translated file but not in the original translation file
- show (option `-p`) or automatically add (`-a`) keys that are not yet translated. Automatically added keys use the original translation marked with a configurable prefix (`--missing-key-prefix`, default: "TRANSLATE")

#### Sample usage

Show (print to console) all missing German translations:
```bash
bin/console mapbender:normalize-translations -p de
```

Automatically add all missing Spanish translations to the yaml files:
```bash
bin/console mapbender:normalize-translations -a es
```

#### Sample output
```bash
Processing translation for Mapbender\CoreBundle  …
Found unexpected key "mb.core.simplesearch.admin.title.help" in translated file not present in original
Translations keys normalized for bundle Mapbender\CoreBundle. No keys were missing.
``

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Features:
 * Added splash screen for all applications ([PR#1522](https://github.com/mapbender/mapbender/pull/1522))
 * [Coordinates Utility](https://github.com/mapbender/coordinates-utility) is now longer a separate repository but integrated as 
   a separate bundle in this repo.
+* New console command `mapbender:normalize-translations` to quickly find and complement missing translations ([PR#1538](https://github.com/mapbender/mapbender/pull/1538))
 * [SearchRouter] New option exportcsv to download the result list as CSV ([PR#1509](https://github.com/mapbender/mapbender/pull/1509))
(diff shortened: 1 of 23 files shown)

### CODE REVIEW

**Comment by astroidex** [comment]
very nice feature. I tried it for it. Worked fine.

What do you think about changing the prefix behaviour?

At the moment it will write Translate: in the text 

```
      label: 'TRANSLATE: Scale'
```

I expected it like this: The missing text would be added and deactivated by # at the beginning. Like that the default language would still be used.

```
# TRANSLATE:      label: 'Scale'
```

But just a suggestion. if you work on the translation in a separate branch and merge it when finished. TRANSLATE: wold not show up in the application.

astroidex (approved):

astroidex:
I will add the inrofmation about the new command to the documentation

Phocacius:
I decided to not modify the keys (i.e. not comment them out) for two reasons:
- the YamlParser just ignores all comments. Therefore, if you'd execute the command a second time without the `-a` flag, these comments would be gone when writing the file back to disc. 
- if the translation keys show up during development, them being marked in the translation string can give additional context to the translator

Of course, the prefixed strings should never be merged (or even committed) into the repository, it's just an intermediate state for the translators.

Issue #1506, opened 2023-09-27
[Layertree]  Smartphone usage with Chrome or Opera: you can not de/activate layer in the layertree.


astroidex:
**Describe the bug**
If you use a Mapbender application (Version 3.3.5) on a Smartphone with Chrome or Opera you can not aktivate/deactivate layer in the layertree.
Everything is fine with firefox.
No problems with older Mapbender Versions
- see  3.3.5 auf https://demo.mapbender.org/application/mapbender_user_yml

**Expected behavior**
Layer handling should work with all browsers

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 3.3.5
 - Browser Chrome / Opera

**Additional context**
Maybe the error is connected to the changes via PR https://github.com/mapbender/mapbender/pull/1457

rschwammborn:
Hello @astroidex ,
we just tested this and could not reproduce the checkbox behavior (Chrome on iOS). Feel free to reopen this if the issue does still persist, maybe with more details concerning the device and operating system used.

astroidex:
Please note that the issue occurs on for example on Google Pixel 4a & 6. So I reopen the ticket.

astroidex:
Error still occurs on with Chrome. I could reproduce it with version 4.0. Should be solved in Verison 4.0

Phocacius:
Fix will be available in the 4.0 release.

Temporary fix for older versions: Replace the file `application/mapbender/src/Mapbender/CoreBundle/Resources/public/jquery-ui-touch-punch/jquery.ui.touch-punch.js` with the contents of this file: https://github.com/RWAP/jquery-ui-touch-punch/blob/master/jquery.ui.touch-punch.js

PR #1539, opened 2024-02-01
Make side pane user-resizable

Phocacius:
Adds an option to the side pane region's properties in the default Fullscreen template:

grafik

It is active by default. The key for yaml applications is `resizable`.

Users can now drag the side pane on its right edge to make it smaller or larger. 

By default, the minimum size is 120px. the maximum is 95% of the screen's width. These values can be further restricted by using custom css:

```css
.sidePane.resizable {
  min-width: 200px;
  max-width: 500px;
}
```


astroidex:
I tested the feature.

- only works with sidepane position left. When you switch to right it does not work
- border of sidepane is bigger when activated. Is that mandatory? or could it only be bigger on mouse over?
- also effect with many buttons see screenshot

sidepane

Phocacius:
- I'd leave it as default. It does affect all existing applications, however it's only an enhancement, it doesn't break existing applications
- Right position is now supported
- border is weakened
- overlap occurs when a single word is very long, maybe add to the documentation that a shy character (https://unicode-explorer.com/c/00AD) can be included in the title to auto-enable hyphenation

astroidex:
works fine

astroidex (approved):

Phocacius:
The update of jquery-ui-touch-punch should also fix #1506, @astroidex please check that again.

astroidex:
works fine with firefox and chrome

PR #1547, opened 2024-02-14
Prevent dragging a popup window outside the viewport

Phocacius:
Until now it was possible to drag a popup window (like print or legend) outside  the viewport by releasing the mouse button outside the current browser window. The dialog was then gone until the respective element was disabled and reenabled.

New restrictions:
- a popup window's left edge may never be placed further left than the left of the viewport
- a popup window's top edge may never be placed further left than the top of the viewport
- a popup window's right edge may never be placed further than 50px from the right edge of the viewport (which leaves the heading visible so it can be dragged back to the main screen)
- a popup window's bottom edge may never be placed further than 50px from the bottom edge of the viewport (which leaves the heading visible so it can be dragged back to the main screen)


astroidex:
ok. Works fine.

astroidex (approved):

PR #1550, opened 2024-02-28
[ImageTransport] Return null if image download fails

### LINKED ISSUE #1549: "500 Internal Server Error" when encountering a network error during printing
**Describe the bug**
If mapbender is setup behind a proxy and a remote WMS-Layer is active (e.g. the OSM Demo layer from mapbender_user) printing fails with a 500 Error. Mapbender tries to get the remote resource as an image, but it does not check if it succeeded. 

In '[src/Mapbender/PrintBundle/Component/Transport/ImageTransport.php](https://github.com/mapbender/mapbender/blob/fe5fa11ef0e2646becd47ee90c47a58076c1d3f2/src/Mapbender/PrintBundle/Component/Transport/ImageTransport.php#L38)' between line 38 and 39 there should be a check like : 
```php
if ($image == false) {
      /*
       * emit_log("failed to create image from remote resource at $url");
       */
      return null;
}
```

(Note: a check like that was removed with "[701e88a:  Update print image download error handling (nothing rows \RuntimeException)](https://github.com/mapbender/mapbender/commit/701e88aaba3d09fd6ef34b4a8738488c2e221db6)"

With this fix applied the 500 Error is fixed, but the resulting PDF file contains only the pdf-template.

**To Reproduce**
1. Setup mapbender and don't allow the server to contact the internet
2. Try to print one of the demo apps
3. Observe either a 500 error or an empty PDF (aside from the template)

**Expected behavior**
Either a correct PDF or an error that mapbender could not fetch a remote resource.

 - Mapbender: 3.3.5
 - Database: postgres-15.6
 - Browser: Not relevant

enno-t:
Currently if downloadImage() is unable to download an image, it tries to do the Gd-Operations on null.

Instead downloadImage() should check if $image is null. With this patch, if $image is indeed null an error gets logged (since a null image is undesirable, but a failed download is not necessarily the wrong behaviour for a system) and returns null. This requires a new argument to downloadImage() ($logger), which all callers have to pass. Most already have a $logger that can be used, but LegendHandler.php does not, so a new normally-null logger is introduced and set from PrintService.

Fixes #1549

### DIFF
--- a/src/Mapbender/PrintBundle/Component/LayerRendererWms.php
+++ b/src/Mapbender/PrintBundle/Component/LayerRendererWms.php
@@ -132,7 +132,7 @@ class LayerRendererWms extends LayerRenderer
         if (count($grid->getTiles()) === 1) {
             // Single-tile grid can trivially be resolved with a single request, avoiding the temporary
             // image used for tile merging.
-            $layerImage = $this->imageTransport->downloadImage($baseUrl, $layerDef['opacity']);
+            $layerImage = $this->imageTransport->downloadImage($baseUrl, $layerDef['opacity'], $this->logger);
         } else {
             $layerImage = imagecreatetruecolor($grid->getWidth(), $grid->getHeight());
             imagesavealpha($layerImage, true);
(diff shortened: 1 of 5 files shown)


astroidex:
@enno-t thanks a lot for the PR. We will review soon.

Phocacius:
Thank you for your fix! I added the functionality via #1551 to the repo. 

For the next PR: Prefer injection over passing services around as method parameters and always add a changelog entry.

PR #1552, opened 2024-02-29
New command mapbender:wms:assign

Phocacius:
#### Description:  
  Assigns a WMS source to an application

#### Usage:
  `mapbender:wms:assign <application> <source> [<layerset>]`

####  Arguments:  
  `application`:              id or slug of the application
  `source`:                      id of the wms source
  `layerset` (optional): id or name of the layerset. Defaults to 'main' or the first layerset in the application.


astroidex:
Tested. Worked fine. Just small note
- error if you pass a WMS id that is not present in the installation. Better raise an info that the WMS-id is not present.

```
  Mapbender\CoreBundle\Component\Source\TypeDirectoryService::createInstance(): Argument #1 ($source) must be of type Mapbender\CoreBundle\Entity\Source, null given, called in /data/mapbender-develop/application/mapbender/src/Mapbender/ManagerBun  
  dle/Controller/SourceInstanceController.php on line 458        
```

astroidex (approved):

PR #1559, opened 2024-03-19
Fix wildcard translation fallback language if not defined in the target language

Phocacius:
Previous (buggy behaviour): If a wildcard translation was defined in an element's required assets and none of the wildcard strings was defined in the target language, the TranslationCompiler threw an error causing no translations at all to be available. 

New behaviour: For wildcard translations the fallback locale's translation is used like it already was the case for non-wildcard translations.


astroidex (approved):

astroidex:
tested works fine

PR #1562, opened 2024-04-08
Russian translation for Mapbender 4

astroidex:
- missing translations are added to the translation files
- translation still has to be done


astroidex (commented):
Thank you Olga for the translation

PR #1565, opened 2024-04-10
Fix visiblelayers param to work by name

PhilVanB:
This enhances the `visiblelayers` url parameter to work with given source- and layer-name also. So far only source- and layer-ID are allowed.
@astroidex Please test if it works as expected.

see documentation to find out about visiblelayers
https://doc.mapbender.org/en/elements/basic/map.html#make-layer-visible

?visiblelayers=Mapbender/Mapbender_Names,Mapbender/Mapbender_User,39/149


astroidex (commented):
I tested the PR. Works fine.

astroidex (approved):

Phocacius (approved):

PR #1566, opened 2024-04-11
Update ImageTransport.php

frwg:
In case the Request fails and returns an Exception, the application used to continue uninformed about the reason of the failure. The proposed change makes debugging a lot easier.


frwg:
I am not sure whether this fits to the general logging/exception handling philosophy, so please adapt it if necessary.

Phocacius (approved):

Issue #1414, opened 2022-04-08
[PrintClient] does not print kartenspiegel on the top as defined in the template. Map is on top instead


astroidex:
**Describe the bug**
In Mapbender 3.0.8.x you could produce a print output with  a kartenspiegl on top of the map as defined in the print template.

You have to produce the pdf as follow
- Strg+A to select all printable elements
- Export as PDF -> Choose Selection
- the print works fine and adds the Kartenspiegel on top
- looks like the pdf has additional information when you choose "Selection" on pdf creation

![pdf_printclient](https://user-images.githubusercontent.com/996298/162414411-e4b6c604-1641-43df-9937-45a3e9709c32.png)

**To Reproduce**
1. Go to PrintClient
2. Create a print pdf
3. See error: map is on top of the Kartenspiegel 

**Expected behavior**
The PrintClient should take into account the ordering and should put the Map-Element at the right position

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: bug in 3.2.x, 3.3 (worked fine in 3.0.8.x)

See pdf / odg attaced to test the print with kartenpiegel
[a4_landscape_kartenspiegel_on_top.pdf](https://github.com/mapbender/mapbender/files/8450842/a4_landscape_kartenspiegel_on_top.pdf)
[a4_landscape_kartenspiegel_on_top.odg](https://github.com/mapbender/mapbender/files/8450843/a4_landscape_kartenspiegel_on_top.odg)

astroidex:
Has to do with the following change in function buildPdf: https://github.com/mapbender/mapbender/commit/367e828ef24c671198582696d1e5f9fc4baf3fe6#diff-03a4c9b662d234d98b9539f6ec141786e7c96fe8efb1e80abbdc0dded43abd6eL202

Works fine with the following code (when the template pdf was generated with PDF-Export -> Selection)
    protected function buildPdf($mapImageName, $templateData, $jobData)
    {
        // @todo: eliminate instance variable $this->pdf
        $this->pdf = $pdf = $this->makeBlankPdf($templateData, $jobData['template']);
        $tplidx = $pdf->importPage(1);

        $this->addMapImage($pdf, $mapImageName, $templateData);
        unlink($mapImageName);

        $pdf->useTemplate($tplidx);

        $this->afterMainMap($pdf, $templateData, $jobData);

        return $pdf;
    }

astroidex:
tested. Fixed in Version 4.0

PR #1568, opened 2024-04-12
[Print Client] Support print of a textbox on top of the map

astroidex:
- some pdf print templates use a textbox on top of the map
- this worked in older version
- see Ticket #1414 - will support textbox on the map again. But needs pdf template generated via selection
- modified function supports the print on top again, but it need a pdf templated hat was generated via Select all -> Export PDF - Selection


swinkelmann:
This PR has been replaced by this one: https://github.com/mapbender/mapbender/pull/1573

PR #1572, opened 2024-04-17
Make searchrouter results sortable

PhilVanB:
The following Searchrouter enhancements were added:

- sort result table by column
- configure initial sort column
- zoom to result extent
- highlight selected table row

Configure initial sort column like so:

- Add `sortBy` and `sortOrder` under result section in yaml-configuration:

```...
searchrouter:
  ...
  results:
     ...
     sortBy: columnName
     sortOrder: asc|desc
     ...
```

Configure zoom to result extent like so:

- Add `zoomToResultExtent` under result section in yaml-configuration:

```...
searchrouter:
  ...
  results:
     ...
     zoomToResultExtent: true
     ...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ Features:
 * [LayerTree] When activating a layer, all its parent layers are also activated ([PR#1544](https://github.com/mapbender/mapbender/pull/1544))
 * [SearchRouter] Extends configuration to handle labeling. ([PR#1553](https://github.com/mapbender/mapbender/pull/1553))
 * [Map] visiblelayers parameter now supports also rootlayer and layer name (not only sourceinstanceid, instanceid) ([PR#1565](https://github.com/mapbender/mapbender/pull/1565))
-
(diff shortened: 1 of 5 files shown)


astroidex:
I tested the feature. Looks very good.
- sorting works fine
- sort via sortBy works fine


Two remarks
- selected Feature in the Table is highligthed  (very good)  -> when you sort via a column -> the selection gets lost
- it is not very obvious that the columns can be sorted - any idea how to make it more obvious?

astroidex (approved):

astroidex:
@PhilVanB and @astroidex discussed

- zoomToResultExtent: true will go to route. To make it configurable per route
- selection should be still there when the sorting is changed
- add icons to table header to see the possibility of sorting

astroidex (changes requested):
remember sorting for next search does not work properly yet.

astroidex (approved):

astroidex (changes requested):
Add Changelog Info

astroidex (approved):

Phocacius (approved):

PR #1573, opened 2024-04-23
Fix/print static items on map

swinkelmann:
- some pdf print templates use a textbox on top of the map
- this worked in older version
- see Ticket #1414 - will support textbox on the map again. But needs pdf template generated via selection
- modified function supports the print on top again, but it need a pdf templated hat was generated via Select all -> Export PDF - Selection
- see discussion in ticket #1414

! This PR replaces https://github.com/mapbender/mapbender/pull/1568


astroidex (approved):
I tested the PR. Works fine.

Phocacius:
@astroidex The "Outlines" directory and its effect should be documented in https://docs.mapbender.org/develop/en/elements/export/printclient.html#create-your-individual-templates

PR #1577, opened 2024-05-02
Fix WMTS form template

swinkelmann:
Fixes the backend form template for WMTS.
The WMTS can then be integrated as a data source again.
However, the WMTS must still be configured as 'activated' when the application is started!


astroidex (changes requested):
Please add a row to the CHANGELOG.md file. PR 1577

astroidex:
I tested the funktionality. Works fine.

Phocacius (approved):

PR #1578, opened 2024-05-02
Digitizer: Change datatables style to boostrap

PhilVanB:
Added bootstrap style for datatables


astroidex:
@Phocacius please check the code. It is conneted to the digitizer PR

astroidex (approved):
ok as I can say when tested with digitizer PR

src/Mapbender/CoreBundle/Asset/ApplicationAssetService.php
```diff
@@ -223,6 +223,7 @@ protected function getFrontendBaseAssets(string $type): array
                     '@MapbenderCoreBundle/Resources/public/init/element-sidepane.js',
                     '@MapbenderCoreBundle/Resources/public/widgets/toolbar-menu.js',
                     '/components/datatables/media/js/jquery.dataTables.min.js',
+                    'components/datatables/media/js/dataTables.bootstrap4.min.js',
```
Phocacius, line 226: does the bootstrap4 template work reliably for us? (we have bootstrap 5 now)

The composer package of datatables is once again quite outdated, the newest available version is 1.10.21 from 10 years ago, the current repo is https://github.com/DataTables/DataTablesSrc/tags

Also, you should probably add a / at the start or remove it from the line above, but I'd have it consistent

PhilVanB: @Phocacius I fixed the assets path. It works reliable, even with datatables' Bootstrap 4 version. You may merge now. Considering the datatables update, I suggest we stick with the old version for now and run the update later on in a second step.

PR #1581, opened 2024-05-15
[Ruler]: Add option to make type user-selectable and allow usage in sidepane

Phocacius:
Added choice "both" for the type option along with "line" and "area" to offer the user the choice to switch between the two modes on the fly.

grafik

Also, using the ruler in the sidepane was not possible before, now both the usage as popup (triggered via a button) and in the sidepane is possible.


astroidex (approved):
works fine, as expected

PR #1582, opened 2024-05-15
[FeatureInfo] Allow usage in side-pane

Phocacius:


astroidex (approved):
works fine. Maybe improve the layout of the results

PR #1584, opened 2024-05-23
[Sketch] Already use selected color while drawing

Phocacius:
Previously, a generic blue (OL default) was used while drawing in the sketch tool, only after finishing the drawing process the actual colour was used. Now, the selected colour is already used while drawing the shape.

Also, added convenience method `Mapbender.StyleUtil.createDrawStyle` that is also used by the ruler element.


astroidex (approved):
works fine

PR #1590, opened 2024-06-04
Fix/minorbugs

swinkelmann:
Fixes some minor bugs:

- Missing dialog title in POI dialog
- don't drag a draggable zoombar outside the screen
- change helptext popover (backend security) behaviour ->trigger on hover instead of click so it closes automatically
- german translation for 'dimension' in layertree config


astroidex:
- OK Missing dialog title in POI dialog
-   ERROR  don't drag a draggable zoombar outside the screen -> when you move the toolbar to the bottom the rest of the applications (sidepane, overview) moves upwards
- change helptext popover (backend security) behaviour ->trigger on hover instead of click so it closes automatically
  - OK as it hides on close (but dialog  does not show up on hover) 
- OK    german translation for 'dimension' in layertree config

PR #1595, opened 2024-06-06
prevent entering overview widget without overview -2

frwg:


frwg:
it's more appropriate to check for the overwiew field on the overviewWidget itself

PR #1596, opened 2024-06-07
allow disabling of dropdown

frwg:


frwg:
dropdown can now be "disabled", i.e. will change it's style and do not respond on click once a "disabled" attribute has been added

PR #1607, opened 2024-07-23
Add ValidatableConfigurationInterface to validate an element's configuration

Phocacius:
An interface `Mapbender\CoreBundle\Component\ElementBase\ValidatableConfigurationInterface` with the method `validate` was added. This enables an element to validate its configuration, e.g. to check that the minimum extent is smaller than the maximum extent in the map element. 

The method is called in two cases:
- when saving a form in the administration backend. The `$form` attribute will be non-null. You should create a form error in this case, e.g. `$form->get('configuration')->get('mykey')->addError(new FormError('Something went wrong'));` 

Example of the error display:

grafik

- when accessing an application in the frontend. In this case, the `$form` argument will be null. This also allows YAML-defined applications to be checked. Throw a `Mapbender\CoreBundle\Component\ElementBase\ValidationFailedException` if a validation error occurs. Caution: This message will be shown to frontend users.

Example of the error display:

grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Bugfixes:
 
 Other:
 * Extract Application Resolving Logic to separate service that can be overwritten by DI ([PR#1604](https://github.com/mapbender/mapbender/pull/1604))
+* Add ValidatableConfigurationInterface to validate an element's configuration ([PR#1607](https://github.com/mapbender/mapbender/pull/1607))
 
 
 ## v4.0.0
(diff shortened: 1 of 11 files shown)


astroidex:
- when there is an error it is shown - fine
- But: When I change f.e. max y to 1 the red error is shown at min y -> should show both columns in red

Translation:
- DE better: call columns directly
  -> min y muss kleiner sein als max y.
  Else you do not directly see the columns involved

Bildschirmfoto vom 2024-07-25 16-31-06

PR #1610, opened 2024-07-24
[DataUpload] Fix data type recognition in some browsers / operating systems

### LINKED ISSUE #1603: [File Upload] GPX and KML uploads fails on OS X
**Describe the bug**
In the file uploading element, the upload fails showing the message "Unsupported file type:" when trying to upload a gpx or kml (maybe also other types) on Mac OS X or iOS

**To Reproduce**
- Use a macbook/imac or iphone (any browser)
- Add a File Upload element to your application
- Upload any .gpx or .kml file to the "upload" element (e.g. from [here](https://www.radroutenplaner-deutschland.de/veraDaten_EN.asp))

**Expected behavior**
The file is uploaded and shown on the map

 - Mapbender: 4.0.0
 - Database: n.a.
 - tested on Firefox and Chrome, but probably any on OS X

**Additional Context:**
The upload element uses the JavaScript `File` API. To determine which data type was uploaded, `File.type` is used. The mdn docs state ([source](https://developer.mozilla.org/en-US/docs/Web/API/Blob/type)): 

``` 
blob.type is generally reliable only for common file types like images, HTML documents, audio and video. 
Uncommon file extensions would return an empty string
```

An alternative way of determining the file type needs to be found.

Phocacius:
fixes #1603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## next bugfix release
 Bugfixes:
 * Do not crash when passport does not exist in FailedLoginListener ([PR#1601](https://github.com/mapbender/mapbender/pull/1601))
+* [DataUpload] Fix data type recognition in some browsers / operating systems ([#1603](https://github.com/mapbender/mapbender/issues/1603), [PR#1610](https://github.com/mapbender/mapbender/pull/1610))
 
 Other:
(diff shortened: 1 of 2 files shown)


astroidex:
Tested
- tested on OS X, iPhone: works fine

astroidex (approved):

PR #1614, opened 2024-07-25
[Security] Fix `published: false` was ignored in YAML applications

Phocacius:
A YAML application was only hidden from public view if the `published` key was not present at all, `published: false` was interpreted like `published: true`. This PR fixes this behaviour.


astroidex (approved):
Tested works fine

PR #1619, opened 2024-08-20
Fix compatibility with OracleDB

cioddi:
Rename field fom_permission.resource to fom.permission.resource-ref to prevent oracledb reserved word issues


astroidex:
@cioddi looks good to me. But I did not test.

Issue #1617, opened 2024-08-08
[Bug] Mobile Template Layertree - scroll is not possible when sort layers is active 


astroidex:
**Describe the bug**
Mobile Template: when "Allow Reorder" for layers is activated in the Layertree it is not possible to scroll. 

**To Reproduce**
1. Go to https://schulung.foss.academy/mapbender4a/application/mapbender_mobile_bug
2. "Allow Reorder" is activated: so you can move wms but you can not scroll to the bottom (on mobile devices)
3. When you deactivate "Allow Reorder" in the backend or can scroll again 

- see also mailing list conversation https://lists.osgeo.org/pipermail/mapbender_users/2024-August/006395.html

**Expected behavior**
it should always be possible to scroll

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 4.0.0
- Browser firefox

Phocacius:
Additional context: Currently the whole width of the layer is a target for the jQuery  UI sortable widget. The options for resolving this issue are:

- restricting the area where reordering is possible. E.g. the touch area where reordering is possible might be restricted to the burger menu at the right side, or the margin on the left side could be used for scrolling while the rest allows reordering
- use two-finger gestures for scrolling while one-finger gestures still only are used for reordering. E.g. Google Maps uses this pattern. This must be communicated though since it's not intuitive.

PR #1620, opened 2024-08-20
[Mobile Template] Allow scrolling layer tree on the left edge

Phocacius:
refs #1617

Solution: Replaced margin on the left hand side of the layer tree by padding. On the left hand side (20px wide), scrolling is now possible while the rest of the row still allows for reordering layers.


astroidex (approved):
Tested
Works as expected. Thank you.

Issue #1618, opened 2024-08-14
Upload images link is wrong when $assetService->registerAssetOverride(); is used


astroidex:
**Describe the bug**
If you use $assetService->registerAssetOverride oder   $assetService->registerAssetOverrides in your own bundle the links to your application upload images are not references right.Upload images are not found.

```
    public function boot(): void
    {
        parent::boot();
        $assetService = $this->container->get('mapbender.application_asset.service');  
        $assetService->registerAssetOverride();
    }
```
![uploads](https://github.com/user-attachments/assets/a8439991-84c0-4b2e-91ce-15b9d6ce403b)

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 4.0.0

PR #1622, opened 2024-08-20
Move asset overriding to its own service to avoid request stack problems

Phocacius:
Problem was, that the request stack is not yet initialised when booting the bundle. However, through many interdependencies, the PathPackage was required in the ApplicationAssetService initialising the basePath to empty string.

Solution: Do not inject the ApplicationAssetService service when registrering overrides in the bundle, but a standalone new service ApplicationAssetOverrides, that does not have any dependencies and therefore also does not trigger the PathPackage to be initialised.

:warning: Any application already using asset overriding needs to be changed:

When injecting the service: `mapbender.application_asset.service` -> `mapbender.application_asset.overrides.service`

When using the class directly: `Mapbender\CoreBundle\Asset\ApplicationAssetService` -> `Mapbender\CoreBundle\Asset\ApplicationAssetOverrides`

refs #1618


astroidex:
- tested. Works fine. Thanks you.

astroidex (approved):

Issue #1624, opened 2024-08-22
[map][visiblelayers] does not activate the root-Layer if only the root layer is passed in the url


astroidex:
**Describe the bug**
does not activate the root-layer
https://schulung.foss.academy/mapbender/application/1_mapbender_user_db?visiblelayers=wms_foss4g/
works fine and activates the root layer too
https://schulung.foss.academy/mapbender/application/1_mapbender_user_db?visiblelayers=wms_foss4g/foss4g

**To Reproduce**

**Expected behavior**
when only the root layer is passed it shoul activate the root layer

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 4.0.1

Phocacius:
In your example application, the root layer is already activated by default

astroidex:
No please check: WMS
- https://schulung.foss.academy/mapbender/application/1_mapbender_user_db?

![Bildschirmfoto vom 2024-08-22 11-03-44](https://github.com/user-attachments/assets/7c90bca7-9c1b-492e-8d82-5069327425e4)

PR #1625, opened 2024-08-22
Allow visiblelayers to activate root layer

Phocacius:
refs #1624


astroidex (approved):
Tested. Works fine with number and name

Issue #1615, opened 2024-07-25
[Map] Gauss-Krueger SRS not working correctly


Phocacius:
**Describe the bug**
When you change the srs to 31466 or 31467 Gauss-Krueger, the map extent is not preserved. Instead, some random place is shown, zooming is not possible

**To Reproduce**
- https://demo.mapbender.org/application/mapbender_user_yml or https://schulung.foss.academy/mapbender4a/application/mapbender_user

1. open application -> Bonn is visible
2. change EPSG to 31466 -> ERROR: very small section is visible

**Expected behavior**
After step 3. the same extent of Bonn should be visible

**Versions (please complete the following information if it helps to narrow down the bug):**
 - Mapbender: 4.0.0

PR #1626, opened 2024-08-26
Fix transformation of EPSG:3847 coordinates

swinkelmann:
Fixes https://github.com/mapbender/mapbender/issues/1615


swinkelmann:
If you switch from EPSG:3847 to UTM or GK, a new max extent is calculated.
Unfortunately, OpenLayers does not calculate a max extent that covers the entire world in UTM or GK.

This is the max extent that is set in mapbender_users:
[ -20037508.34278924, -20037508.34278924, 20037508.34278924, 20037508.34278924 ]
After the transformation to UTM32 we get this:
[ -586331.6298902088, -10543719.766426057, 586331.6298902095, 10543719.766426057 ]
So we only get a 1700 km wide strip as max extent. That's why you can't pan further east.

Possible solutions:
- We don't set a worldwide max extent in the mapbender_users application.
- We don't allow switching to UTM or GK

Phocacius:
Merging now as it's the best solution we have at the moment. If we find a better solution in the future it will be part of the 4.0.2 release.

PR #1631, opened 2024-09-10
Feature/delete dialog backgroundinformation2

dholler1011:
Added backgroundinformation to popups "delete user" and "delete layout". Updated all available translations.
see internal ticket #6543


astroidex:
Hello @dholler1011 delete user and element works fine. Please improve your PR and form proper sentences via using variables in the translation text. See my changes for delete user and please improve delete element and also the changes for the missing languages.

src/FOM/UserBundle/Resources/translations/messages.de.yaml
```diff
@@ -67,7 +67,7 @@ fom:
         edit_user: 'Benutzer bearbeiten'
         contact_user: 'Kontaktieren Sie den Benutzer per E-Mail'
         edit_group: 'Gruppe bearbeiten'
-        delete_user: 'Benutzer löschen'
+        delete_user: 'Möchten Sie den Benutzer %user_username% wirklich löschen'
```
Phocacius, line 70: Please add quotes around the template parameter, so it later renders as `Möchten Sie den Benutzer "the user" wirklich löschen`. That way it is visually easier to see it is a parameter.

src/Mapbender/ManagerBundle/Resources/views/Application/form-elements.html.twig
```diff
@@ -88,7 +88,7 @@
                     <i class="fas fa-key"></i>
                 </span>
                 <span class="-fn-delete hover-highlight-effect"
-                      title="{{ 'mb.actions.delete' | trans }}"
+                      title="{{"mb.manager.admin.element.delete.title"|trans}} {{ element_title(element) | trans }}"
```
Phocacius, line 91: As Astrid already said: Please add a template parameter here as well

Phocacius (changes requested):

src/Mapbender/ManagerBundle/Resources/views/Application/form-elements.html.twig
dholler1011: The translations of the text for deleting the user were no problem. But the text for deleting an element could not be translated easily. The latter only worked after the translation text was moved to another yaml-file. Furthermore, I could not find any other examples in the source code in which texts from e.g. application/mapbender/src/Mapbender/ManagerBundle/Resources/translations/messages+intl-icu.de.yaml, which are composed of several parts in a twig-template, are correctly translated or applied. 
Do you have any idea why this is the case?

Phocacius (approved):

astroidex (approved):

PR #1633, opened 2024-09-18
sort-function backend

dholler1011:
added sort-function for users, usergroups and global authorizations
see internal ticket #7369

### DIFF
--- a/src/Mapbender/ManagerBundle/Resources/public/components.js
+++ b/src/Mapbender/ManagerBundle/Resources/public/components.js
@@ -22,38 +22,32 @@ $(function() {
         return false;
     });
 
-    $(document).on('click', ".sortByColumn", function(){
+    $(document).on('click', ".sortByColumn", function () {
         var $this = $(this);
         var asc = true;
 
-        $this.css('text-decoration', 'underline');
-        $this.siblings().css('text-decoration', 'none');
-        $this.siblings().find('span').hide();
+        $this.siblings().find('span').removeClass().addClass('fa fa-sort').prop('title', Mapbender.trans('mb.actions.sort_ascending'));
 
-        if($this.find('span').is(":hidden")){
-            $this.find('span').removeClass('fa-arrow-down').addClass('fa-arrow-up').show();
-        }else if($this.find('span').hasClass('fa-arrow-up')){
+        if ($this.find('span').hasClass('fa fa-sort-amount-asc')) {
             asc = false;
-            $this.find('span').removeClass('fa-arrow-up').addClass('fa-arrow-down');
-            $this.prop('title', Mapbender.trans('mb.actions.sort_ascending'));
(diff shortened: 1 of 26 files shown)


astroidex:
- sorting works

Please implement the following enhancements
- but please take the jquery table style used in the searchRouter  (ask @PhilVanB is you have questions) to have the same style
- please open the list sorted by user name

**jquery table style looks like this**
searchrouter_sort

- see code: https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Resources/public/mapbender.element.searchRouter.js#L305

src/Mapbender/ManagerBundle/Resources/public/components.js
```diff
@@ -22,6 +22,45 @@ $(function() {
...
+    $(document).on('click', ".sortByColumn", function(){
+        var $this = $(this);
+        var asc = true;
+
+        $this.css('text-decoration', 'underline');
+        $this.siblings().css('text-decoration', 'none');
+        $this.siblings().find('span').hide();
+
+        if($this.find('span').is(":hidden")){
+            $this.find('span').removeClass('fa-arrow-down').addClass('fa-arrow-up').show();
```
Phocacius, line 34: Similar to my comment in #1632, here design and functionality is very tightly coupled which makes customisations more complicated. A solution to decouple it would be to have two spans instead of one with the classes `sort-indicator-asc` and `sort-indicator-desc` (or similar, in addition to the fa-classes) and only toggle their visibility. That way only css and/or the twig template needs to be modified to change the appearence.

src/Mapbender/ManagerBundle/Resources/public/components.js
```diff
@@ -22,6 +22,45 @@ $(function() {
...
+        }else if($this.find('span').hasClass('fa-arrow-up')){
```
Phocacius, line 35: And please run the auto-formatter once before creating a PR, on some places (like here) the space is missing after the brace. This avoids a line showing as modified later when an unrelated change is made.

Phocacius (changes requested):
Functionality works as expected, thanks! I have two remarks:

src/Mapbender/ManagerBundle/Resources/public/components.js
dholler1011: I actually have to toggle between 3 classes: `fa fa-sort`, `fa fa-sort-amount-asc` and `fa fa-sort-amount-desc`. I couldn't think of a technically nice solution how to toggle with 3 classes (or 3 spans) without making the source code more complicated.

Phocacius: alright :+1:

Phocacius (approved):

astroidex (approved):

PR #1634, opened 2024-09-23
Fix ajax requests fail in YAML applications (regression introduced in 4.0.1)

### LINKED ISSUE #1628: Requests to ElementController do not work with YML applications
All requests to Mapbender elements do not work with YML applications since version 4.0.1.

Example:
- Open the mapbender_user (YML) application
- Open the network tab in browser DevTools
- Open the print client
-> no rectangle appears on the map
-> you cannot switch the templates
-> the template request (../application/mapbender_user/element/printclient/getTemplateSize?template=a4landscape) ends with 404

Reason:
A small change was made in the ElementController in this commit: 882a0e5597c987e845831ef5eae0b8d5ca6e8e10
`$id = intval($id);` 
This conversion is okay when using database applications because $id is numeric.
In YML applications $id is a string like 'printclient'. This ends in an error.

Possible solution:
We could remove the intval conversion again. 
But first we should ask @Phocacius what its actual purpose is. Maybe security reasons? (sql injection)

Phocacius:
closes #1628

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## next bugfix release
+Bugfixes:
+* Fix ajax requests fail in YAML applications (regression introduced in 4.0.1) ([#1628](https://github.com/mapbender/mapbender/issues/1628), [PR#1634](https://github.com/mapbender/mapbender/pull/1634))
+
+
 ## v4.0.1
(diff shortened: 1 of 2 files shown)


swinkelmann (approved):
That looks good. :thumbsup:

PR #1636, opened 2024-10-07
Feature/switching styles via context menu in the layertree

dholler1011:
Add selectbox to the contextmenu in the layertree if a WMS has several selectable styles
see internal ticket 7956


Phocacius (changes requested):
Code looks good, during testing I encountered a WMS that did not work though (demo application via private message)

Phocacius (approved):

PR #1657, opened 2024-11-25
[PrintClient] Various improvements in LegendHandler 

Phocacius:
### New method `measureLegendBlock`
This is mostly improving developer experience. The measuring of a legend block (title, image and overall width/height) and the actual PDF rendering are separated. This was implemented for a client project, where the legend should only be printed on the map page if it fits on the page entirely.

### Correctly handle multi-line headings
The title width calculation now respects the page margins.

grafik
Left: before, right: after

### Smaller default margins between legend blocks
Until now, the default margins were very big, creating a huge unnecessary amount of whitespace on the legend page. This was reduced.

grafik
Left: before, right: after

### Automatically adjust graphic size when it's too big for a column
If a single legend graphic was too big even if it's the only image in  a column it was cropped off. Now, it gets scaled down to fit on the page.

grafik
Left: before, right: after

### Add option for dynamic column widths
Column widths in the legend were fixed until now (variable `$maxColumnWidthMm` within the LegendHandler). This PR adds a new option `dynamicColumnSizes` to the LegendHandler that adds a non-default option to adapt column widths depending on the content.

grafik
Left: dynamicColumnSizes = false (widths fixed to 10cm), Right: dynamicColumnSizes = true

To enable this feature, you need to overwrite the LegendHandler in your client project:

`parameters.yaml`:
```yaml
parameters:
   mapbender.print.legend_handler.service.class: App\Component\Print\LegendHandler
```

`LegendHandler.php`:
```php
<?php

namespace App\Component\Print;

class LegendHandler extends \Mapbender\PrintBundle\Component\LegendHandler
{

    public function __construct(
        #[Autowire(service: 'mapbender.imageexport.image_transport.service')] ImageTransport $imageTransport,
        #[Autowire('mapbender.print.resource_dir')] string                                   $resourceDir,
        #[Autowire('mapbender.print.temp_dir')] ?string                                      $tmpDir,
        #[Autowire('mapbender.print.canvas_legend.class')] ?string                           $canvasLegendClass
    )
    {
        parent::__construct($imageTransport, $resourceDir, $tmpDir, $canvasLegendClass);
        $this->dynamicColumnSizes = true;
    }
}
```

### Make margin between legend title and graphic configurable
Until now, there was no gap between legend title and legend graphic. This is now configurable in the LegendHandler.

grafik
Left: no margin (old behaviour), right: 2mm margin between title and graphic.

Adjusting this feature also requires overwriting the LegendHandler, specifically the array key `title_to_image` within the `getMargins` method:

`parameters.yaml`:
```yaml
parameters:
   mapbender.print.legend_handler.service.class: App\Component\Print\LegendHandler
```

`LegendHandler.php`:
```php
<?php

namespace App\Component\Print;

use Mapbender\PrintBundle\Component\TemplateRegion;

class LegendHandler extends \Mapbender\PrintBundle\Component\LegendHandler
{
    protected function getMargins(TemplateRegion $region): array
    {
        return [
            ...parent::getMargins($region),
            'title_to_image' => 2, //mm
        ];
    }
}
```

### DIFF
--- a/src/Mapbender/PrintBundle/Component/Legend/LegendBlock.php
+++ b/src/Mapbender/PrintBundle/Component/Legend/LegendBlock.php
@@ -14,7 +14,7 @@ class LegendBlock extends GdCanvas implements LegendBlockContainer
     protected $rendered = false;
 
     /**
-     * @param resource $image GDish
+     * @param \GdImage $image GDish
      * @param string $title
      */
     public function __construct($image, $title)
--- a/src/Mapbender/PrintBundle/Component/LegendHandler.php
+++ b/src/Mapbender/PrintBundle/Component/LegendHandler.php
@@ -9,6 +9,7 @@ use Mapbender\PrintBundle\Component\Legend\LegendBlockGroup;
 use Mapbender\PrintBundle\Component\Pdf\PdfUtil;
 use Mapbender\PrintBundle\Component\Region\FontStyle;
 use Mapbender\PrintBundle\Component\Region\FullPage;
+use Mapbender\PrintBundle\Component\Region\NullRegion;
 use Mapbender\PrintBundle\Component\Transport\ImageTransport;
 
 /**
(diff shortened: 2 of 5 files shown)


astroidex:
I tested the functionality. Works fine. i will add an example to the workshop bundle.

astroidex (approved):

PR #1662, opened 2025-01-13
Refactor Wmts to use ES6 classes

Phocacius:
Addition to #1635, WMTS sources were not yet refactored


PhilVanB (approved):
This works as expected now.

PR #1665, opened 2025-01-14
[GpsPosition] Disable checkbox zoomToAccuracyOnFirstPosition when centerOnFirstPosition is not activated

Phocacius:
see internal ticket 4230


astroidex:
- Tested
- backend checkbox behaviour works fine

astroidex (approved):

astroidex:
Shall we add info to changelog?

Phocacius:
I thought it was too minor to add it to the changelog. We can do it though if you think it's worth it.

astroidex:
yes. Too minor. So it is ok like that.

PR #1666, opened 2025-01-14
[ApplicationSwitcher] Allow reordering applications by drag and drop in manager

Phocacius:
see internal ticket 8653


astroidex:
Tested, works fine.

astroidex (approved):

PR #1669, opened 2025-01-15
[SearchRouter] Compare extents to features in EPSG:4326 only

Phocacius:
Per default, the map extent is now compared to a feature by converting both to EPSG:4326 (WGS84).
Comparing in a non-global SRS (like UTM32) can result in errors when comparing it to an extent defined in a global CRS. This may reduce performance, so the transformation to EPSG:4326 can be disabled by passing `noTransform:true` to `class_options`.

see internal ticket 9356


astroidex:
- Tested. Works fine. Performance looks ok.
- with noTransform: true the result has errors when comparing it to an extent defined in a global CRS.

astroidex (approved):

PR #1670, opened 2025-01-15
Support data-label attribute in feature info highlighting

Phocacius:
In addition to data-geometry and data-srid the new attribute data-label is read from a feature in the feature info html (QGIS: maptip). 

It can be styled using the new options 
- `fontColorDefault` (default: #000000)
- `fontColorHover` (default: #000000)
- `fontSizeDefault` (default: 12)
- `fontSizeHover` (default: 12)

see internal ticket 7433


astroidex:
I tested the new feature
works fine.

Phocacius:
@astroidex Please add to documentation.

PR #1672, opened 2025-01-17
[API] Add new ApiBundle

swinkelmann:
Adds the ApiBundle with endpoints required by the QGIS2Mapbender plugin

see internal ticket 9226

se also https://github.com/mapbender/mapbender-starter/pull/134 for details


src/Mapbender/ApiBundle/Controller/ApiController.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Mapbender\ApiBundle\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Attribute\Route;
+use OpenApi\Attributes as OA;
+
+class ApiController extends AbstractController
```
Phocacius, line 11: do we still need this class? I would say it's not necessary anymore, but I'd also be fine with keeping it. In this case though,  I'd also add the denyAccessUnlessGranted here

src/Mapbender/ApiBundle/Controller/CommandController.php
```diff
@@ -0,0 +1,324 @@
...
+        responses: [
+            new OA\Response(
+                response: 200,
+                description: 'Success message with WMS data',
+            )
+        ]
+    )]
+    public function prepareWmsShowCommand(Request $request, KernelInterface $kernel): JsonResponse
+    {
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_ACCESS_API);
```
Phocacius, line 51: This works, but unfortunately outputs a HTML message again, which is not practical for an API. Similar to the unsuccessful login, this should return something like `{"code": 403, "message": "Access deniced"}`

src/Mapbender/ApiBundle/Controller/CommandController.php
```diff
@@ -0,0 +1,324 @@
...
+            new OA\Response(
+                response: 200,
+                description: 'Success message',
+            )
+        ]
+    )]
+    public function prepareWmsAssignCommand(Request $request, KernelInterface $kernel): JsonResponse
+    {
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_ACCESS_API);
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_EDIT_ALL_APPLICATIONS);
```
Phocacius, line 257: hm, I'd say edit rights on the one specific application would be sufficient to execute this. This would however require injecting the ApplicationResolver to obtain the application instance.

src/Mapbender/ApiBundle/Controller/CommandController.php
```diff
@@ -0,0 +1,324 @@
...
+            new OA\Response(
+                response: 200,
+                description: 'Success message with the new slug',
+            )
+        ]
+    )]
+    public function prepareApplicationCloneCommand(Request $request, KernelInterface $kernel): JsonResponse
+    {
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_ACCESS_API);
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_CREATE_APPLICATIONS);
```
Phocacius, line 306: can we assume that someone who has CREATE_APPLICATION rights has also read rights on the existing application? If not, this should also be checked

src/Mapbender/ApiBundle/Controller/UploadController.php
```diff
@@ -0,0 +1,104 @@
...
+use OpenApi\Attributes as OA;
+
+class UploadController extends AbstractController
+{
+    private string $uploadDir;
+
+    public function __construct(string $uploadDir)
+    {
+        $this->uploadDir = $uploadDir;
+    }
```
Phocacius, line 20: Since we have PHP8.1 minimum requirement now, you can make your life easier :)

`public function __construct(private string $uploadDir) {}` is enough.

src/Mapbender/ApiBundle/Controller/UploadController.php
```diff
@@ -0,0 +1,104 @@
...
+    )]
+    public function uploadZipAction(Request $request): Response
+    {
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_ACCESS_API);
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_UPLOAD_FILES);
+
+        $file = $request->files->get('file');
+
+        if (!$file) {
+            return new Response('No file uploaded', Response::HTTP_BAD_REQUEST);
```
Phocacius, line 69: I would keep the format consistent, i.e. return an object with `success: true` and `message` as the message (applices to all Responses in this method)

src/Mapbender/ApiBundle/Controller/CommandController.php
```diff
@@ -0,0 +1,427 @@
...
+        if (!$this->isGranted(ResourceDomainInstallation::ACTION_ACCESS_API)) {
+            $missingPermissions[] = 'Access API';
+        }
+        if (!$this->isGranted(ResourceDomainInstallation::ACTION_CREATE_APPLICATIONS)) {
+            $missingPermissions[] = 'Create Applications';
+        }
+
+        $appEntity = $this->applicationResolver->getApplicationEntity($slug);
+        $this->denyAccessUnlessGranted(ResourceDomainApplication::ACTION_VIEW, $appEntity);
+        if (!$this->isGranted(ResourceDomainApplication::ACTION_VIEW)) {
```
Phocacius, line 379: Die beiden Zeilen müssten noch gemergt mergen, also

```php
if (!$this->isGranted(ResourceDomainApplication::ACTION_VIEW, $appEntity)) {
```

Ansonsten sieht von meiner Seite aus alles gut aus!

Phocacius (approved):

PR #1679, opened 2025-01-27
[Routing] Fix transport type, Round off coordinates

PhilVanB:
- Fixes the transport type selection 
- Rounds off the coordinates to 2 digits after comma


astroidex (changes requested):
klick start/ klick end
- route should be shown directly

add an intermediate point
- rout should be updated immediately

PhilVanB:
- auto-submit should now be working on any changes
- precision for the coordinate values is now insured

Please check

astroidex (approved):
I tested again. Works fine

PR #1682, opened 2025-01-30
Fix order of elements in application config

swinkelmann:
This error occurs when an element is deleted in an application. The order of the elements is then changed in ALL other applications.

The only thing missing was a where clause for the current application.

see also internal ticket 9071


astroidex (approved):
I approve by viewing the code, not testing as it is hard to reproduce.

Phocacius (approved):

PR #1689, opened 2025-02-06
Include upgrade of old button class (Mapbender\CoreBundle\Element\Button) to mapbender:database:upgrade command

Phocacius:


swinkelmann (approved):
If the string 'http' is in the group or tooltip, a link button would be created here by mistake. But this should not actually happen.

approved 👍

astroidex (approved):

PR #1698, opened 2025-02-25
[ShareUrl] Fix selected layers not encoded correctly

### LINKED ISSUE #1697: Share Url generated a link that is not activating the layers that were selected
**Describe the bug**
You can activate layer and zoom to a region and then create a link via Share link. The link now contains object object instead of a number or name. The link does not activate the layer.

https://xxxx/mapbender/application/mapbender_user_db?slon=977%3A%5Bobject%20Object%5D#5810@7.09977/50.73399r0@EPSG:3857

**Expected behavior**
The link should contain the name of the layer that should be activated

 - Mapbender: 4.1

Phocacius:
fixes #1697


astroidex (approved):
works fine now

astroidex (approved):

PR #1700, opened 2025-03-04
WMS styles in View manager and share url state

Phocacius:
This PR saves the custom selected WMS style when saving ViewManager states and when sharing a url via the ShareUrl element.


astroidex (approved):
- works fine
- testes share url and ViewManager with Styles

Issue #1701, opened 2025-03-07
[SimpleSearch] Backend: Validation error message not translated


StBamberger:
If you type wrong inputs in the configuration form for the SimpleSearch Element in the backend and try to save, an error message is being shown. This error message is not translated an schows only the YAML-Path to the translation.

As this error also occurred when developing my own element based on the source code of SimpleSearch element, I think this issue applies to all Mapbender elements wich use the Symfony Constraints in conjunction with the message parameter set to a YAML key. I solved the issue by using the trans() method of the translator interface to explicitly translate the text. But I think this should be fixed more centralized, so that not all elements have to be checked and changed if necessary.
Eventually this issue also applies to other funtions using the Symfony Contraints interface.

1. Add a new SimpleSearch Element to the application layout
2. Set the correct settings for a search configuration
3. Now change one validated value to a erroneous value, for example the value for the source SRS to "foo"
4. Click Apply button
5. the message under the inut field shows "mb.core.simplesearch.errors.invalid_epsg_code" instead of the translation

Expected behavior:
The translated text from the language file is shown

Versions:
 - Mapbender: 4.1.0
 - Browser: Chrome 133.0

Source hint:
./mapbender/src/Mapbender/CoreBundle/Element/Type/SimpleSearchAdminConfigurationType.php Line 129

Phocacius:
Thank your for your report! The issue was fixed and will be included in the next bugfix release.

For your own elements: The translation for Constraint messages needs to be placed in the file `validators.<countr>.yaml` instead of `messages.<country>.yaml`

PR #1702, opened 2025-03-11
Fix validation messages for simple search were in the wrong translation file

Phocacius:
refs #1701


momotosushi (approved):
Übersetzungen funktionieren einwandfrei

PR #1705, opened 2025-03-14
Cloning of elements enabled

dholler1011:
Elements in Layout-lists can be cloned by clicking on copy-icon;
see ticket 7592

### DIFF
--- a/src/Mapbender/ManagerBundle/Controller/ElementController.php
+++ b/src/Mapbender/ManagerBundle/Controller/ElementController.php
@@ -428,6 +428,43 @@ class ElementController extends ApplicationControllerBase
     }
 
     /**
+     * Duplicate element
+     *
+     * @param Request $request
+     * @param string $slug
+     * @param string $id
+     * @return Response
+     */
+    #[ManagerRoute('/application/{slug}/element/{id}/duplicate', name: 'mapbender_manager_element_duplicate', methods: ['POST'])]
+    public function duplicateAction(Request $request, $id)
+    {
+        $this->denyAccessUnlessGranted('ROLE_ADMIN');
+
+        $element = $this->getRepository()->find($id);
+        $application = $element->getApplication();
+
+        if ($this->isCsrfTokenValid('element_duplicate', $request->request->get('token')) === false) {
+            throw new InvalidCsrfTokenException('Invalid CSRF token.');
+        }
+
+        try {
+            $clonedElement = clone $element;
+
+            $sameRegionCriteria = Criteria::create()->where(Criteria::expr()->eq('region', $element->getRegion()));
+            $regionSiblings = $application->getElements()->matching($sameRegionCriteria);
+            $newWeight = $regionSiblings->count();
+            $clonedElement->setWeight($newWeight);
+
+            $this->em->persist($clonedElement);
+            $this->em->flush();
+
+            return new Response('', 200);
+        } catch (\Exception $e) {
+            return new Response($e->getMessage(), 500);
+        }
+    }
+
+    /**
      * @return EntityRepository
      */
     protected function getRepository()
--- a/src/Mapbender/ManagerBundle/Controller/ElementController.php
+++ b/src/Mapbender/ManagerBundle/Controller/ElementController.php
@@ -427,28 +427,33 @@ class ElementController extends ApplicationControllerBase
         return new Response(null, Response::HTTP_NO_CONTENT);
     }
 
-    /**
-     * Duplicate element
-     *
-     * @param Request $request
-     * @param string $slug
-     * @param string $id
-     * @return Response
-     */
     #[ManagerRoute('/application/{slug}/element/{id}/duplicate', name: 'mapbender_manager_element_duplicate', methods: ['POST'])]
-    public function duplicateAction(Request $request, $id)
+    public function duplicateAction(Request $request, $slug, $id): Response
     {
-        $this->denyAccessUnlessGranted('ROLE_ADMIN');
-
         $element = $this->getRepository()->find($id);
+
+        if (!$element) {
+            throw $this->createNotFoundException("The element with the id \"$id\" does not exist.");
+        }
+
         $application = $element->getApplication();
+        $this->denyAccessUnlessGranted(ResourceDomainApplication::ACTION_EDIT, $application);
 
         if ($this->isCsrfTokenValid('element_duplicate', $request->request->get('token')) === false) {
-            throw new InvalidCsrfTokenException('Invalid CSRF token.');
+            $this->addFlash('error', $this->translator->trans('mb.manager.admin.csrf_token_invalid'));
+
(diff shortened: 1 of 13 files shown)


astroidex (changes requested):
Hello @dholler1011, thanks for the PR. Looks very good. Here are some remarks for discussion.

1. please use icon far fa-copy (as it is used for example to copy applications) too. 
2. do not allow to copy some elements, as it will cause to produce an invalid application (for example map-element, maybe also overview, others??)
3. cloning element with restrictions (key) looses the restrictions 
4. when you clone an element (for example legend) and after that you want to add a button to connect with the duplictaed element the element (in this case legend) is only shown once in the target-selection (element with title legend is there twice). When you change the title for the copied element it shows up in the target-selection. -> solution could be to provide a title for a copied element? f.e. legend-1

Phocacius:
I would argue against the restriction for some elements (your comment 2).  If you want to try tweaking some settings without losing the original settings, it's a valid use case to duplicate the element, disable the first, tweak your settings and eventually delete one of the two instances.

astroidex:
comment 2: ok. Fine. Maybe we should add also some notes in the copy dialog that the user might have to modify the copied element or might break the applicaton

src/Mapbender/ManagerBundle/Controller/ElementController.php
```diff
@@ -427,6 +427,43 @@ public function screentype(Request $request, Element $element)
         return new Response(null, Response::HTTP_NO_CONTENT);
     }
 
+    /**
+     * Duplicate element
+     *
+     * @param Request $request
+     * @param string $slug
+     * @param string $id
+     * @return Response
```
Phocacius, line 436: You can add this as native return type now (`public function duplicateAction(Request $request, $id): Response`). The verbose and informationless PHPDoc can be removed then.

src/Mapbender/ManagerBundle/Controller/ElementController.php
```diff
@@ -427,6 +427,43 @@ public function screentype(Request $request, Element $element)
...
+     */
+    #[ManagerRoute('/application/{slug}/element/{id}/duplicate', name: 'mapbender_manager_element_duplicate', methods: ['POST'])]
+    public function duplicateAction(Request $request, $id)
+    {
+        $this->denyAccessUnlessGranted('ROLE_ADMIN');
```
Phocacius, line 441: Requiring admin privileges is too hard of a restriction here. Anyone who can edit the application should also be able to duplicate elements. This is also used in the other methods of this controller, e.g. `new` or `edit`.

```php
$element = $this->getRepository()->find($id);
$application = $element->getApplication();
$this->denyAccessUnlessGranted(ResourceDomainApplication::ACTION_EDIT, $application);
```

src/Mapbender/ManagerBundle/Controller/ElementController.php
```diff
@@ -427,6 +427,43 @@ public function screentype(Request $request, Element $element)
...
+
+        $element = $this->getRepository()->find($id);
+        $application = $element->getApplication();
+
+        if ($this->isCsrfTokenValid('element_duplicate', $request->request->get('token')) === false) {
+            throw new InvalidCsrfTokenException('Invalid CSRF token.');
```
Phocacius, line 447: please use the (already present) translated version: `mb.manager.admin.csrf_token_invalid`

src/Mapbender/ManagerBundle/Controller/ElementController.php
```diff
@@ -427,6 +427,43 @@ public function screentype(Request $request, Element $element)
...
+     * @param string $slug
+     * @param string $id
+     * @return Response
+     */
+    #[ManagerRoute('/application/{slug}/element/{id}/duplicate', name: 'mapbender_manager_element_duplicate', methods: ['POST'])]
+    public function duplicateAction(Request $request, $id)
+    {
+        $this->denyAccessUnlessGranted('ROLE_ADMIN');
+
+        $element = $this->getRepository()->find($id);
```
Phocacius, line 443: Please add a check if the element exists, like in https://github.com/mapbender/mapbender/blob/2e36ce12506a52eac8f6486b45ed1234f0aaccc2/src/Mapbender/ManagerBundle/Controller/ElementController.php#L214

src/Mapbender/ManagerBundle/Controller/ElementController.php
```diff
@@ -427,6 +427,43 @@ public function screentype(Request $request, Element $element)
...
+
+        try {
+            $clonedElement = clone $element;
+
+            $sameRegionCriteria = Criteria::create()->where(Criteria::expr()->eq('region', $element->getRegion()));
+            $regionSiblings = $application->getElements()->matching($sameRegionCriteria);
+            $newWeight = $regionSiblings->count();
+            $clonedElement->setWeight($newWeight);
+
+            $this->em->persist($clonedElement);
```
Phocacius, line 458: As Astrid already pointed out, the cloning of existing permissions for this element is missing here. You can use `PermissionManager::findPermissions` to find the permissions that are present for the existing element and clone them as well.

src/Mapbender/ManagerBundle/Resources/public/components.js
```diff
@@ -422,6 +422,41 @@ $(function() {
...
+                            url: $el.data('url'),
+                            type: 'POST',
+                            data: {
+                                token: $el.data('token')
+                            },
+                            success: function() {
+                                window.location.reload();
+                            },
+                            error: function(xhr) {
+                                Mapbender.error(xhr.responseText || 'Ein Fehler ist aufgetreten.');
```
Phocacius, line 445: Please translate this message. You can use `mb.application.save.failure.general` or create your own message.

src/Mapbender/ManagerBundle/Resources/views/Application/form-elements.html.twig
```diff
@@ -99,7 +99,14 @@
...
                 </span>
-                  {% set element_name = element_title(element) | trans %}
+                {% set element_name = element_title(element) | trans %}
+                <span class="duplicateElement hover-highlight-effect"
+                      title="{{ 'mb.manager.components.popup.duplicate_element.title'|trans}}"
+                      data-title="{{ 'mb.manager.components.popup.duplicate_element.confirm'|trans({'element_name': element_name})}}"
+                      data-slug="{{ application.slug }}"
+                      data-token="{{ csrf_token('element_duplicate') }}"
+                      data-url="{{ path('mapbender_manager_element_duplicate', {'slug': application.slug, 'id': element.id}) }}"
+                      ><i class="fas fa-copy"></i></span>
```
Phocacius, line 109: Like @astroidex said: please use icon far fa-copy (as it is used for example to copy applications) too.

Phocacius (changes requested):

Phocacius (approved):
Thank you, looks perfect now :+1:

PR #1708, opened 2025-04-01
[Print] Allow adding more options for quality levels

Phocacius:


astroidex (approved):
- definitions from before are still there
- allows to add new quality levels
- allows to delete all quality levels - no errors in the client default is used for printing
- when only one quality level is defined it is not shown in the client

PR #1718, opened 2025-04-24
[WMS-Loader] Fix layer in map not loaded correctly

PhilVanB:
Fixes incorrect layer active states in map after loading WMS with WMS-Loader-Links.
-> see Internal #10302


astroidex (changes requested):
- please check again
- 1. When you load a WMS via link at first all layers are requested (although you onls activated one layer)
- 2. when you add another WMS via link with merge and activate another layer -> the change is visible in the layertree, but not in the map

PhilVanB:
@astroidex Please test again, if everything works as expected now.

PhilVanB:
@astroidex root layer is now deactivated, when no layer is given. Please check,

astroidex:
I tested. Works fine. 
- see https://schulung.foss.academy/mapbender/application/mapbender_user_db?#66641@7.09543/50.73512r0@EPSG:25832

astroidex (approved):

Phocacius (approved):

PR #1719, opened 2025-04-29
[ViewManager] Fixes deletion of public views

PhilVanB:
Fixes Internal #10248
Public views can now be deleted by any user, not just the creator.


astroidex (approved):
- tested works fine
- maybe PR to master instead develop
- feature for 4.1.1

PR #1720, opened 2025-05-06
[ApiBundle] Add new options for wms:assign

PhilVanB:
- Adds new options to the mapbender:wms:assign command and the  /api/wms/assign api-request 
- see internal #10243

New options are
- --format
- --infoformat
- --proxy
- --tiled
- --layerorder

get information via

`bin/console mapbender:wms:assign --help`

### DIFF
--- a/src/Mapbender/ApiBundle/Controller/CommandController.php
+++ b/src/Mapbender/ApiBundle/Controller/CommandController.php
@@ -283,6 +283,41 @@ class CommandController extends AbstractController
                 in: 'query',
                 required: false,
                 schema: new OA\Schema(type: 'mixed', example: "main")
+            ),
+            new OA\Parameter(
+                name: 'format',
+                description: 'Sets the format for the GetMap-request, such as image/png',
+                in: 'query',
+                required: false,
+                schema: new OA\Schema(type: 'string', example: 'image/png')
+            ),
+            new OA\Parameter(
+                name: 'infoformat',
+                description: 'Sets the format for the FeatureInfo, such as text/html',
+                in: 'query',
+                required: false,
+                schema: new OA\Schema(type: 'string', example: 'text/html')
+            ),
+            new OA\Parameter(
+                name: 'proxy',
+                description: 'Decides if a proxy is used or not (one of true|false). Defaults to "false"',
+                in: 'query',
+                required: false,
+                schema: new OA\Schema(type: 'boolean', example: true)
+            ),
+            new OA\Parameter(
+                name: 'tiled',
+                description: 'Decides if the GetMap-requests are returned tiled or not (one of true|false). Defaults to "false"',
+                in: 'query',
+                required: false,
+                schema: new OA\Schema(type: 'boolean', example: true)
+            ),
+            new OA\Parameter(
+                name: 'layerorder',
+                description: 'Sets the layerorder to either standard or reverse (one of standard|reverse). Defaults to "standard"',
+                in: 'query',
+                required: false,
+                schema: new OA\Schema(type: 'string', example: 'standard')
             )
         ],
         responses: [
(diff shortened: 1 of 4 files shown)


src/Mapbender/ApiBundle/Controller/CommandController.php
```diff
@@ -283,6 +283,41 @@ public function prepareWmsAddCommand(Request $request, KernelInterface $kernel):
...
+            new OA\Parameter(
+                name: 'format',
+                description: 'Sets the format for the GetMap-request, such as image/png',
+                in: 'query',
+                required: false,
+                schema: new OA\Schema(type: 'string', example: 'image/png')
+            ),
+            new OA\Parameter(
+                name: 'infoformat',
+                description: 'Sets the format for the FeatureInfo, such as text/html',
```
astroidex, line 296: please write FeatureInfo-request

src/Mapbender/ApiBundle/Controller/CommandController.php
```diff
@@ -283,6 +283,41 @@ public function prepareWmsAddCommand(Request $request, KernelInterface $kernel):
...
+            new OA\Parameter(
+                name: 'tiled',
+                description: 'Decides if the GetMap-requests are returned tiled or not (one of true|false). Defaults to "false"',
+                in: 'query',
+                required: false,
+                schema: new OA\Schema(type: 'boolean', example: true)
+            ),
+            new OA\Parameter(
+                name: 'layerorder',
+                description: 'Sets the layerorder to either standard or reverse (one of standard|reverse). Defaults to "standard"',
```
astroidex, line 317: Question: is it not the definition that is defined in the variable **wms.default_layer_order**?

The varibale is per default **wms.default_layer_order: standard**

But it can be defined in the parameters.yaml for example to qgis. The qgis should be used by the mapbender:wms:add command, right?

`wms.default_layer_order: qgis `

src/Mapbender/WmsBundle/Command/SourceAssignCommand.php
```diff
@@ -39,6 +40,11 @@ protected function configure(): void
             ->addArgument(self::ARGUMENT_APPLICATION, InputArgument::REQUIRED, "id or slug of the application")
             ->addArgument(self::ARGUMENT_SOURCE, InputArgument::REQUIRED, "id of the wms source")
             ->addArgument(self::ARGUMENT_LAYERSET, InputArgument::OPTIONAL, "id or name of the layerset. Defaults to 'main' or the first layerset in the application.")
+            ->addOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Sets the format for the GetMap-request, such as image/png', null)
+            ->addOption('infoformat', 'i', InputOption::VALUE_OPTIONAL, 'Sets the format for the FeatureInfo, such as text/html', null)
```
astroidex, line 44: see commend above

astroidex (changes requested):

src/Mapbender/ApiBundle/Controller/CommandController.php
PhilVanB: **wms.default_layer_order** (located in _parameters.yaml_) defines the layer order when adding a new WMS.

On top of that, a single WMS can be configured individually (regarding the layerorder) by specifying `--layerorder`. If this individual configuration is not carried out, the value from _parameters.yaml_ is used.

AnnikaFroede:
Info: I tested with bin/console mapbender:wms:assign

Is it intended, that there is no error, if an option is invalid? 
At the moment it is then just handled as providing the default value.
Example: -p truuue will be handled as 'false', but there is no error or warning that shows me that this input was incorrect

AnnikaFroede:
I get an error, when I try to use the API /api/doc:
image

If I try it with bin/console mapbender:wms:assign with the same application name, it works just fine. This seems like a bug.

Phocacius:
@PhilVanB @AnnikaFroede I fixed the problem with the "Try it out" feature. It worked before using e.g. curl, only the swagger docs complained when using the mixed type.

Phocacius (approved):

PR #1721, opened 2025-05-07
[Routing] Fix routing in popup-dialog

PhilVanB:
Fixes no route and marker displayed when search is repeated. -> Happens when routing is configured in the popup-dialog.
see internal issue #10311


Phocacius (approved):

astroidex (commented):
Tested. Works fine.

PR #1728, opened 2025-06-04
[API] Fix controller response status code

swinkelmann:
Another small fix here, because we cannot (always) use the console command's exit code as the controller's response code. This only works if the console command returns 404, which happens in some cases.

### DIFF
--- a/src/Mapbender/ApiBundle/Controller/CommandController.php
+++ b/src/Mapbender/ApiBundle/Controller/CommandController.php
@@ -461,7 +461,8 @@ class CommandController extends AbstractController
                 return new JsonResponse([
                     'success' => false,
                     'error' => $commandOutput,
-                ], 500);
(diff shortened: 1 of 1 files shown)


src/Mapbender/ApiBundle/Controller/CommandController.php
```diff
@@ -461,7 +461,7 @@ function executeCommand($inputArgs, $kernel)
                 return new JsonResponse([
                     'success' => false,
                     'error' => $commandOutput,
-                ], $exitCode);
+                ], 500);
```
Phocacius: why not check if the response is a valid status code and use it if it exists?

```php
array_key_exists($exitCode, Response::$statusTexts) ? $exitCode : Response::HTTP_INTERNAL_SERVER_ERROR
```

swinkelmann: Okay, let's do it like this :-)

PR #1749, opened 2025-07-18
Layertree filter function

swinkelmann:
This adds an input field to filter the layertree.

See internal ticket 10202

- activate Layertree filter in Element Layertree via checkbox 

layertree_filter

See https://doc.mapbender.org/en/elements/basic/layertree.html

 Modify highlight color via css
```
 .mb-element-layertree .layer-highlight {
  color: #279ddc;
  font-weight: bold;
}
```

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
@@ -98,10 +98,7 @@
             this.element.on('click', '.layer-metadata', function (evt) {
                 self._showMetadata(evt);
             });
-            this.element.on('input', '#layerFilter', this._filterLayer.bind(this));
-            this.element.on('click', '#resetLayerFilter', function (evt) {
-                self._resetLayerFilter(true);
-            });
+            this.element.on('input', '.layer-filter-input', this._filterLayer.bind(this));
             $(document).bind('mbmapsourceloadstart', $.proxy(self._onSourceLoadStart, self));
             $(document).bind('mbmapsourceloadend', $.proxy(self._onSourceLoadEnd, self));
             $(document).bind('mbmapsourceloaderror', $.proxy(self._onSourceLoadError, self));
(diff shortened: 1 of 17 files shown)


src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -98,6 +98,10 @@
             this.element.on('click', '.layer-metadata', function (evt) {
                 self._showMetadata(evt);
             });
+            this.element.on('input', '#layerFilter', this._filterLayer.bind(this));
```
Phocacius: Please don't use ids when referencing DOM elements within an element. An id should be unique by html specification, and there can always be multiple instances of an element within an application (e.g. sidebar for desktop usage and toolbar button + popup for mobile usage)

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -454,6 +458,49 @@
             var layer = $target.closest('li.leave').data('layer');
             this.model.controlLayer(layer, null, newState);
         },
+
+        _resetLayerFilter: function (clearInput) {
+            if (clearInput) $('#layerFilter').val('');
```
Phocacius: all of the jquery calls within this method should also be prefixed `this.element.find( ... )`, also because there might be multiple layertree instances within one application

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -454,6 +458,49 @@
...
+            $('span.filtered').removeClass('filtered');
+            $('span.layer-title').parent().show();
+            $('.themeContainer, .serviceContainer').removeClass('showLeaves');
```
Phocacius: I don't like the behaviour of collapsing all folders when a search is removed. I see two UX-friendlier options:

- Save the collapse/expand state before the first search character is entered and restore it after removing the search string
- (my preferred option and should also be easier to implement): Keep the collapse/expand state as it was during the search, just also show all other layers

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -454,6 +458,49 @@
...
+            $('i.fa-folder-open').removeClass('fa-folder-open').addClass('fa-folder');
+        },
+        _filterLayer: function () {
+            const value = $('#layerFilter').val().toLowerCase();
```
Phocacius: also prefix with this.element and replace id by class

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -454,6 +458,49 @@
...
+
+            if (typeof this._lastFilterLength === 'undefined') {
+                this._lastFilterLength = 0;
+            }
+
+            if (value.length > 2) {
```
Phocacius: I'd discourage having super long ifs if (höhö) they can be avoided. 
In this case here, the _lastFilterLength assignment can be made before the if. Then, work with early returns:

```js
if (value.length < 3) {
   if (this._lastFilterLength > 2) this._resetLayerFilter(false);
   return;
}

// the code from the if clause, without indentation
```

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -454,6 +458,49 @@
...
+                const $layerTitles = $('span.layer-title');
+                $layerTitles.each(function () {
```
Phocacius: I'm also not a fan of reassigning "this" within a class. ALternatively this could be written as

```js
$layerTitles.each((index, element) => {
  const title = $(element).attr('title')  ...
```

(just a suggestion though, IMO code is more readable if "this" is always the same)

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -454,6 +458,49 @@
...
+                        $(this).toggleClass('filtered', title.includes(value));
+                    }
+                });
+
+                $layerTitles.parent().hide();
+                $('span.filtered').each(function () {
+                    $(this).parent().show();
+                    $(this).siblings('span.-fn-toggle-children').find('i').addClass('fa-folder-open');
+                    ['li.serviceContainer', 'li.themeContainer'].forEach(selector => {
+                        const $container = $(this).parents(selector);
```
Phocacius: here it's even trickier, since "this" is the iterator of 'span.filtered' ... takes a while to understand what's going on. I'd vote for using the arrow syntax in the top .each as well.

src/Mapbender/CoreBundle/Resources/public/sass/element/layertree.scss
```diff
@@ -60,6 +60,11 @@ $textColor: $darkFontColor !default;
     opacity: 0;
     cursor: default;
   }
+
+  .btn-outline-secondary:not(:hover) {
```
Phocacius: no need for additional css here, just add the `btn-light` class to the button in the twig template

src/Mapbender/CoreBundle/Resources/translations/messages.de.yaml
```diff
@@ -275,6 +275,9 @@ mb:
         select_style: 'Stil auswählen'
         metadata: Metadaten
         dimension: Dimension
+      filter:
+        placeholder: 'Ebenen filtern...'
```
Phocacius: Please use the `…` character instead of three dots (better e.g. for screenreaders)

src/Mapbender/CoreBundle/Resources/views/Element/layertree.html.twig
```diff
@@ -1,3 +1,11 @@
+{% if configuration.showFilter %}
+    <div class="input-group mb-3">
+        <input type="text" id ="layerFilter" class="form-control" placeholder="{{ 'mb.core.layertree.filter.placeholder'|trans }}" tabindex="0">
+        <button id="resetLayerFilter" class="btn btn-outline-secondary" type="button" title="{{ 'mb.core.layertree.filter.reset'|trans }}">
```
Phocacius: add `btn-light` here (see above). And change all the ids to classes

src/Mapbender/CoreBundle/Resources/public/mapbender.element.layertree.tree.js
```diff
@@ -454,6 +455,70 @@
             var layer = $target.closest('li.leave').data('layer');
             this.model.controlLayer(layer, null, newState);
         },
+        _resetLayerFilter: function () {
+            $(this.element).find('span.layer-highlight').each((index, span) => {
```
Phocacius: (nothing that needs to be addressed, also it does not change anything, just for the future): this.element in a MApbender element widget is already a jQuery element, so wrapping in $() is not necessary.

Phocacius (approved):

astroidex:
```
  .layer-filter-input {
       margin: 3px;
  } 
```
Was added with https://github.com/mapbender/mapbender/pull/1749/commits/3f409cd5e6cc8f9c09f9e986bc75d16ebeb7ba2e

else the border is  not displayed in sidepane and dialog completely

Suche_Layertree

astroidex (approved):
Tested. Works fine. With the last commit you can search for WMS groups and you can still reach the sublayer.

PR #1751, opened 2025-07-25
Work/accessibility optimize keyboard operation testing

dholler1011:
Supplement to PR #1742 :
* Magnifying glass still visible when focused
* Accessibility colorpicker in sketch-element
See internal ticket 10948


Phocacius:
Thanks, works really well! Only thing, I replaced the deprecated `e.keyCode` references by `e.key` which also makes the code more readable.

PR #1755, opened 2025-07-31
Feature/view manager enhancements

PhilVanB:
- saves and recalls the complete layertree status

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/element/mbViewManager.js
+++ b/src/Mapbender/CoreBundle/Resources/public/element/mbViewManager.js
@@ -1,4 +1,4 @@
-;!(function($) {
+(function($) {
     "use strict";
     $.widget("mapbender.mbViewManager", {
         options: {
@@ -140,11 +140,20 @@
             $.when(listingPromise, this.mapPromise)
                 .then(function(response) {
                     var $content = $(response[0]);
+                    const params = Mapbender.Util.getUrlQueryParams(window.location.href);
+                    const viewid = params.hasOwnProperty('viewid') ? params.viewid : false;
+                    let loadViewButton = false;
                     $('a.-fn-apply', $content).each(function() {
                         self._updateLinkUrl(this);
+                        if (parseInt(viewid) === parseInt($(this).closest('tr').attr('data-id'))) {
+                            loadViewButton = $(this);
+                        }
                     });
                     $loadingPlaceholder.replaceWith($content);
                     self._updatePlaceholder();
+                    if (loadViewButton !== false) {
+                        loadViewButton.trigger('click');
+                    }
                 }, function() {
                     $loadingPlaceholder.hide()
                 })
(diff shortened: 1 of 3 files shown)


src/Mapbender/CoreBundle/Resources/public/element/mbViewManager.js
```diff
@@ -140,11 +142,20 @@
             $.when(listingPromise, this.mapPromise)
                 .then(function(response) {
                     var $content = $(response[0]);
+                    const params = Mapbender.Util.getUrlQueryParams(window.location.href);
+                    const viewid = params.hasOwnProperty('viewid') ? params.viewid : false;
```
Phocacius, line 149: When opened via viewid GET param, the view is correctly restored, but the check boxes in the layertree are wrong

src/Mapbender/CoreBundle/Resources/public/element/mbViewManager.js
```diff
@@ -394,11 +421,35 @@
...
         },
-        _apply: function(diff) {
-            var settings = this.mbMap.getModel().mergeSettings(this.referenceSettings, diff);
+        _apply: function(settings) {
+            const layertreeElement = $('.mb-element-layertree');
+            layertreeElement.find('ul.layers:first').empty();
+            this.mbMap.getModel().sourceTree = [];
+            const self = this;
+            this.mbMap.map.olMap.getAllLayers().forEach(layer => {
+                self.mbMap.map.olMap.removeLayer(layer);
```
Phocacius: no need for `self` here, you can just use `this` when using lambda functions

src/Mapbender/CoreBundle/Resources/public/element/resetView.js
```diff
@@ -16,17 +15,23 @@
...
 
         run: function() {
+            let settings = window.localStorage.getItem('viewManagerSettings');
+            settings = JSON.parse(settings);
+            settings = {
+                layersets: JSON.parse(settings.layersetStates),
+                sources: JSON.parse(settings.sourcesStates),
+                viewParams: this.mbMap.getModel().decodeViewParams(settings.viewParams)
+            };
+            $('.mb-element-viewmanager').data('mapbenderMbViewManager')._apply(settings);
```
Phocacius: The "ResetView" element will now no longer work if no ViewManager element is active within the application. Maybe move `saveDefaultSettings` and `_apply` into the mapbender model or the map element?

Also, by convention, methods starting with an underscore are private. If we say it's fine that the ResetView element breaks without a ViewManager, I'd rename the function to not start with an underscore.

src/Mapbender/CoreBundle/Resources/public/element/mbViewManager.js
PhilVanB: This should be fixed now. [Take a look here](https://github.com/mapbender/mapbender/pull/1755/commits/a728b30c4eac5b188d63f50a914279ce418d9f4d)

src/Mapbender/CoreBundle/Resources/public/element/mbViewManager.js
PhilVanB: [Fixed.](https://github.com/mapbender/mapbender/pull/1755/commits/64c4256148e648e26fa5e3b334009e879e4dd60f)

src/Mapbender/CoreBundle/Resources/public/element/resetView.js
PhilVanB: [Fixed.](https://github.com/mapbender/mapbender/pull/1755/commits/2931754e38a0b8daec05bd8eed8c246949361d8f)

src/Mapbender/CoreBundle/Resources/public/element/mbViewManager.js
```diff
@@ -101,16 +105,17 @@
         },
         _initEvents: function() {
             var self = this;
+            this.element.on('mousedown', '.-fn-apply', function() {
+                $(this).closest('tr').find('#loadingspinner').removeClass('d-none');
```
Phocacius: the loading spinner only appears when loading the new view via the plus button, not when clicking anywhere on the table row. Also, it should be a class instead of an id (there can be multiple view managers across the application)

PhilVanB: [Fixed.](https://github.com/mapbender/mapbender/pull/1755/commits/f032fcec8864e554f342fc8792af15d8c4ecf197)

Phocacius (approved):

PR #1757, opened 2025-08-13
Fix WMS from yaml applications and wms loader was no longer working

Phocacius:
fixes regression introduced by #1756


astroidex (approved):
WMS Loader works fine

PR #1762, opened 2025-08-14
Draft: [Legend] Dynamic (viewport-relative) legend for services that support it

Phocacius:
if the new option `dynamicLegend` is activated, a wms or wmts layer's legend is refreshed after each pan or zoom of the map. 

The GetLegendGraphic request will have the additional appended parameters `BBOX` and `CRS`. Servers that support it (e.g. [QGIS Server](https://docs.qgis.org/3.40/en/docs/server_manual/services/wms.html#wms-getlegendgraphic-bbox)) will render only the layers that are visible in the current extent (or the print bbox).

Also, the legend graphics now show a loading spinner while the images are pending.

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.element.legend.js
@@ -201,9 +201,12 @@
         },
         _prepareDynamicLegendParameter: function () {
             const model = this.mbMap.getModel();
+            const imageSize = this.mbMap.getModel().getCurrentViewportSize()
             return {
                 CRS: model.getCurrentProjectionCode(),
                 BBOX: model.getCurrentExtentArray().join(','),
+                WIDTH: imageSize.width,
+                HEIGHT: imageSize.height,
             };
         },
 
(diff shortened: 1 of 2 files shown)


astroidex (changes requested):
in the mapbender_user application (mapserver WMS) I get an error for the legend, as width and height are missing

Error

`msWMSLoadGetMapParams(): WMS server error. Missing required parameter WIDTH `

- ERROR https://wms.wheregroup.com/cgi-bin/mapbender_user.xml?version=1.3.0&service=WMS&request=GetLegendGraphic&sld_version=1.1.0&layer=Mapbender_User&format=image%2Fpng&STYLE=default&CRS=EPSG%3A3857&BBOX=786603.0236790808%2C6579571.918077659%2C820456.5236790808%2C6597591.8540151585&
when I add with and height the legend comes back -no error
- https://wms.wheregroup.com/cgi-bin/mapbender_user.xml?version=1.3.0&service=WMS&request=GetLegendGraphic&sld_version=1.1.0&layer=Mapbender_User&format=image%2Fpng&STYLE=default&CRS=EPSG%3A3857&BBOX=786603.0236790808%2C6579571.918077659%2C820456.5236790808%2C6597591.8540151585&&WIDTH=1968&HEIGHT=1534

Question
- problem is that Mapbender requests a bigger map than the one you can see in the client. How is that handed?
- Question: how are gekkachelte WMS handled?

PR #1782, opened 2025-09-24
[Draft] [Design] Mobile view behaviour

dholler1011:
see internal ticket 10675

### DIFF
--- a/src/Mapbender/PrintBundle/Resources/public/element/printclient.scss
+++ b/src/Mapbender/PrintBundle/Resources/public/element/printclient.scss
@@ -1,7 +1,7 @@
 .mb-element-printclient {
   overflow: visible;
   // Reduce vertical size of potentially very long form
-  font-size: 11px;
+  font-size: 0.9rem;
 
   .button.active {
     background-color: #538e24;


src/Mapbender/CoreBundle/Resources/public/elements/MapbenderElement.js
```diff
@@ -23,10 +23,12 @@ class MapbenderElement {
         };
     }
 
-    activateByButton(callback) {
+    activateByButton(callback, mbButton) {
         this.callback = callback ? callback : null;
+        let $mbButton = mbButton || {};
```
Phocacius: This is not a working fallback. When the mbButton is not supplied, it will become an empty object here and crash in `$mbButton.find('i').attr('class')` (which effectively will be `undefined.attr('class')`.

I'd suggest to call the variable `$mbButton` in the first place and just add a `if ($mbButton)` before settings options.icon.

src/Mapbender/CoreBundle/Resources/public/elements/MapbenderElement.js
```diff
@@ -23,10 +23,12 @@ class MapbenderElement {
...
         if (!this.popup || !this.popup.$element) {
             const options = this.getPopupOptions();
+            options.icon = $mbButton.find('i').attr('class') || $mbButton.find('span.mb-glyphicon').attr('class') || '';
```
Phocacius: As also stated in #1789, the fact that the button is a i or span element is an implementation detail that should not be relied upon. Maybe the new class attribute in findIcon from there can also be used to add a class like `js-mb-icon`, that can be used here.

src/Mapbender/CoreBundle/Resources/public/widgets/mapbender.popup.js
```diff
@@ -3,6 +3,7 @@
     window.Mapbender = window.Mapbender || {};
 
     var currentModal_ = null;
+    var mobileBreakpoint = 599; // Mobile breakpoint in pixels
```
Phocacius: This should be configurable, and as a scoped variable within here it is not. Better add it as a class variable in the Popup constructur (`this.mobileBreakpoint = 599`), that way it can be easily changed when overriding the class. Also, add a comment in _popup.scss that the breakpoint needs to be changed here too, otherwise that will be forgotten very easily.

src/Mapbender/CoreBundle/Resources/public/widgets/mapbender.popup.js
```diff
@@ -189,6 +198,14 @@
                 this.$modalWrap = null;
             }
         },
+        icon: function(icon) {
```
Phocacius, line 208: please auto-format this, the spacing is not consistent

src/Mapbender/CoreBundle/Resources/public/widgets/mapbender.popup.js
```diff
@@ -238,12 +255,59 @@
...
+
+            // Mobile resize functionality
+            this.setupMobileResize_($target);
+        },
+        setupMobileResize_: function($target) {
+            var self = this;
+            var isDragging = false;
+            var startY = 0;
+            var startHeight = 0;
+            var minHeight = 200;
```
Phocacius: IMO this should also be configurable (like the mobileBreakpoint)

src/Mapbender/CoreBundle/Resources/public/widgets/mapbender.popup.js
```diff
@@ -238,12 +255,59 @@
...
+            var maxHeight = window.innerHeight - 100;
```
Phocacius: why 100?

src/Mapbender/CoreBundle/Resources/public/widgets/mapbender.popup.js
```diff
@@ -238,12 +262,57 @@
...
+                $('body').addClass('popup-resizing').css('user-select', 'none');
+            });
+
+            $(document).on('mousemove touchmove', function(evt) {
+                if (!isDragging) return;
+
+                evt.preventDefault();
+                var currentY = evt.type === 'touchmove' ? evt.originalEvent.touches[0].clientY : evt.clientY;
+                var deltaY = currentY - startY;
+                var newHeight = Math.max(self.mobileResizeMinHeight, Math.min(self.mobileResizeMaxHeight, startHeight + deltaY));
```
Phocacius: one more problem here: When dragging the popup very far down so the slider disappears, there's no way to make the popup smaller again. A solution for this problem is to further limit the maximum height based on the clientHeight minus the position of the popup element, since the popup does not start on the very top of the page, but below the toolbar. 

Code for this:

```js
const newHeight = Math.max(
    self.mobileResizeMinHeight,
    Math.min(
        self.mobileResizeMaxHeight,
        startHeight + deltaY,
        window.innerHeight - self.$element[0].getBoundingClientRect().y
    )
);
```

src/Mapbender/CoreBundle/Resources/public/elements/MapbenderElement.js
Phocacius: postponed for now, may be tackled later

PR #1783, opened 2025-09-30
[WMTS] Check if current projection is supported by WMTS layer and show message if it isn't

Phocacius:
Can be tested by https://osm-demo.wheregroup.com/wmts/1.0.0/WMTSCapabilities.xml
It supports Pseudo-Mercator but not UTM32. When trying to load in UTM32 is now shows a message in the layertree:

grafik


astroidex (approved):
Tested. Works fine.

PR #1825, opened 2026-02-16
Check access rights in instance tunnel controller

Phocacius:
Follow-up-PR to #1812: Instead of just not displaying the shared instance when it's secured, also check the access rights in
 /application/{slug}/instance/{instanceId}/tunnel route


PhilVanB (approved):
Approved. Works as expected.

PR #1831, opened 2026-03-10
[OgcApiFeatures] Add new OgcApiFeatures source

PhilVanB:
See Pull [#165](https://github.com/mapbender/mapbender-starter/pull/165)

Needs database update: `php bin/console doctrine:schema:update --complete --force`

YAML-Config for OGC-API-Features-Source:

 ```YAML
 liegenschaften_nrw:
      type: ogc_api_features
      title: Liegenschaftskataster NRW
      jsonUrl:  https://ogc-api.nrw.de/lika/v1
      layers:
        31: { collectionId: flurstueck, title: Flurstück, active: true, minScale: ~, maxScale: ~, selected: true, allowSelected: true, info: true, allowInfo: true, featureLimit: 90 }
        32: { collectionId: flurstueck_punkt, title: Flurstückspunkt, active: true, minScale: ~, maxScale: ~, selected: true, allowSelected: true, info: true, allowInfo: true, featureLimit: 90 }
        30: { collectionId: gebaeude_bauwerk, title: 'Gebäude, Bauwerk', active: true, minScale: ~, maxScale: ~, selected: true, allowSelected: true, info: true, allowInfo: true, featureLimit: 90 }
      visible: false
      opacity: 100
      isBaseSource: false
      selected: false
      allowSelected: true
      toggle: false
      allowToggle: true
      minScale: ~
      maxScale: ~
      featureLimit: 100
      featureInfoPropertyMap:
        -
          aktualit: 'Aktualität'
        -
          'gebnutzbez'
```

### CODE REVIEW

File: src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesLoader.php [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+        if (preg_match('#/v(\d+(?:\.\d+)*)#', $url, $matches)) {
+            $version = 'v' . $matches[1];
+        }
+
+        $baseUrl = str_replace('/collections', '', $url);
+        $source->setJsonUrl($baseUrl);
+        $source->setTitle($json['title'] ?? 'Ogc Api Features Source');
+        $source->setDescription($json['description'] ?? '');
+        $source->setVersion($version);
+        $source->setAttribution($json['attribution']);
```
Phocacius: attribution seems to be optional, it does not exist on https://schulung.foss.academy/cgi-bin/mapserv/freiburg/ogcapi. 


```suggestion
        $source->setAttribution($json['attribution'] ?? '');
```

src/Mapbender/OgcApiFeaturesBundle/Entity/OgcApiFeaturesSource.php
```diff
@@ -0,0 +1,97 @@
...
+namespace Mapbender\OgcApiFeaturesBundle\Entity;
+
+use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ORM\Mapping as ORM;
+use Mapbender\CoreBundle\Entity\Source;
+use Mapbender\OgcApiFeaturesBundle\OgcApiFeaturesDataSource;
+
+#[ORM\Entity]
+#[ORM\Table(name: 'mb_ogc_api_features_source')]
```
Phocacius, line 12: the discriminator column mb_core_source::discr has a limit of 15 chars, `ogc_api_features` has 16 resulting in an SQL error when inserting. Should update the limit in https://github.com/mapbender/mapbender/blob/develop/src/Mapbender/CoreBundle/Entity/Source.php#L19

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+            'treeOptions' => [
+                'selected' => $sourceInstance->getSelected(),
+                'toggle' => $sourceInstance->getToggle(),
+                'allow' => [
+                    'selected' => $sourceInstance->getAllowSelected(),
+                    'toggle' => $sourceInstance->getAllowToggle(),
+                ],
+            ],
+            'featureInfo' => [
+                'propertyMap' => $featureInfoPropertyMapExists ? json_decode($sourceInstance->getFeatureInfoPropertyMap(), true) : null,
```
Phocacius: that can be cached above to avoid having to do the decoding twice

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+        ];
+        $config['state'] = [
+            'info' => $this->featureInfoEnabled($sourceInstance),
+        ];
+
+        foreach (array_reverse($sourceInstance->getLayers()->toArray()) as $layer) {
+            if ($layer->getActive()) {
+                $config['children'][] = [
+                    'options' => [
+                        'id' => $layer->getId() . '_',
```
Phocacius, line 63: why the additional underscore?

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+                        'priority' => $layer->getPriority(),
+                        'title' => $layer->getTitle(),
+                        'collectionId' => $layer->getSourceItem()->getCollectionId(),
+                        'minScale' => (!empty($layer->getMinScale()) ? $layer->getMinScale() : $sourceInstance->getMinScale()),
```
Phocacius: a value of 0 is also `empty` in PHP. Shouldn't 0 be allowed to potentially override the default setting of the instance?

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+        return $config;
+    }
+
+    protected function featureInfoEnabled($sourceInstance): bool
+    {
+        $featureInfoEnabled = false;
+        foreach ($sourceInstance->getLayers() as $layer) {
+            if ($layer->getInfo() === true) {
+                $featureInfoEnabled = $layer->getInfo();
+                break;
```
Phocacius: you can just `return true` here and remove the variable entirely

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesInstanceFactory.php
```diff
@@ -0,0 +1,157 @@
...
+        $instance->setSource($source);
+        $instance->setTitle($data['title']);
+        $instance->setOpacity($data['opacity']);
+        $instance->setAllowSelected($data['allowSelected']);
+        $instance->setSelected($data['selected']);
+        $instance->setAllowToggle($data['allowToggle']);
+        $instance->setToggle($data['toggle']);
+        $instance->setMinScale($data['minScale']);
+        $instance->setMaxScale($data['maxScale']);
+        $instance->setFeatureLimit($data['featureLimit']);
```
Phocacius: These values should all have `?? xy` fallbacks if they're not defined (same for the layers below)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesInstanceFactory.php
```diff
@@ -0,0 +1,157 @@
...
+                        $match = $existingLayer;
+                        break;
+                    }
+                }
+                if ($match) {
+                    $entityPool->add($match, $layerIdentData);
+                } else {
+                    return false;
+                }
+            }
```
Phocacius, line 145: This can lead to unexpected behaviours. If in the DB there's a source with layers A and B and from YAML a source with layers A and C is imported, the layer A will already be added to the entity pool, layer C can't be found though, so it and the source itself are not added. A solution would be to cache $match and $layerIdentData into an array and add them to the EntityPool only after the loop has successfully ended without a return.

src/Mapbender/OgcApiFeaturesBundle/Resources/views/edit-instance.html.twig
```diff
@@ -0,0 +1,84 @@
...
+    <script>
+        $('.collectionTable').each(function() {
+            $('tbody', this).sortable({
+                cursor: 'move',
+                axis: 'y',
+                items: 'tr',
+                distance: 6,
+                containment: 'parent',
+                stop: (event, ui) => {
+                    $('input[type="hidden"]', $('.collectionTable tbody tr')).each(function(idx, item) {
```
Phocacius: maybe a bit nit-picky, but if there will ever be another hidden field added, this will break. A more concise approach to get the priority is to check if the id ends with "priotrity"
```suggestion
                    $('input[id$="priority"]', $('.collectionTable tbody tr')).each(function(idx, item) {
```

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesRenderer.php
```diff
@@ -0,0 +1,21 @@
...
+use Mapbender\PrintBundle\Component\Export\Box;
+use Mapbender\PrintBundle\Component\Export\ExportCanvas;
+use Mapbender\PrintBundle\Component\Export\Resolution;
+use Mapbender\PrintBundle\Component\LayerRenderer;
+
+class OgcApiFeaturesRenderer extends LayerRenderer
+{
+    public function addLayer(ExportCanvas $canvas, array $layerDef, Box $extent, array $jobData): void
+    {
+        // No need for implementation so far - print works
```
Phocacius: maybe add to the comment that the print works because the features are added as an OpenLayers VectorSource, and those are automatically added to the print via MbImageExport::_collectGeometryLayers

src/Mapbender/OgcApiFeaturesBundle/Resources/views/edit-instance.html.twig
```diff
@@ -0,0 +1,84 @@
...
+                <td>{{ form_widget(layer.info) }}</td>
+            </tr>
+        {% endfor %}
+        </tbody>
+    </table>
+{% endblock %}
+
+{% block js %}
+    {{ parent() }}
+    <script>
```
Phocacius, line 88: for the help popover to work, you also need to add the initialisation js, see https://github.com/mapbender/mapbender/blob/master/src/Mapbender/VectorTilesBundle/Resources/views/edit-instance.html.twig#L40

src/Mapbender/OgcApiFeaturesBundle/Resources/public/geosource.ogc_api_features.source.js
```diff
@@ -0,0 +1,181 @@
...
+        this.nativeLayers = [];
+        this._vectorSources = {};
+        this.getRootLayer().children.forEach((child) => {
+            const collectionId = child.options.collectionId;
+            this._currentRequestIds[collectionId] = 0;
+            const vectorSource = new ol.source.Vector({
+                strategy: ol.loadingstrategy.bbox,
+                loader: (extent, resolution, projection) => {
+                    this._loadCollection(collectionId, extent, resolution, projection, vectorSource);
+                }
```
Phocacius, line 19: The loading indicators are not yet working. They listen to the `mbmapsourceloadstart` / `error` / `end` events. For WMS, this is done globally using https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Resources/public/mapbender.model.js#L190, when using a custom loader, I'd assume the events should be created manually.

src/Mapbender/OgcApiFeaturesBundle/Resources/public/geosource.ogc_api_features.source.js
```diff
@@ -0,0 +1,181 @@
...
+            this.nativeLayers.push(vectorLayer);
+        });
+        this.setOpacity(this.options.opacity);
+        return this.nativeLayers;
+    }
+
+    _loadCollection(collectionId, extent, resolution, projection, source) {
+        const requestId = ++this._currentRequestIds[collectionId];
+        const projCode = projection.getCode();
+        const bboxExtent = ol.proj.transformExtent(extent, projCode, 'EPSG:4326');
```
Phocacius, line 36: (Just a question): Is it guaranteed by the specification that the data is always in WGS 84?

src/Mapbender/OgcApiFeaturesBundle/Resources/public/geosource.ogc_api_features.source.js
```diff
@@ -0,0 +1,181 @@
...
+                console.error(`Failed to load collection "${collectionId}":`, err);
+            });
+    }
+
+    getSelected() {
+        return this.getRootLayer().getSelected();
+    }
+
+    updateEngine() {
+        const children = this.getRootLayer().children;
```
Phocacius, line 74: When reordering the sources in the frontend, it behaves a bit weird. Sometimes nothing happens, sometimes reordering happens even though only deselected layers have been moved, see video (sorry for the wrong resolution when recording):

https://github.com/user-attachments/assets/460ffcdf-f974-4aa8-a69d-44decd0559f8

Phocacius (changes requested):

astroidex (changes requested):
I could not load the service and got an SQL error. 
- solution:I had to modify the column disc in Table mb_core_source as text "ogc_api_features" has length 17 characters
- suggestion change the size from varchar(15) to varchar(50)
- BaseSourceswitcher: Feature Instance can be set to Basesource - Then it can be selected in the Element BaseSourceSwitcher -> But in the Application the new BaseSource is not visible

1. Symbol same as WMS
2. add possibility to toggle all on/off (see WMS)
3. smaller input 
4. Title mor space for input
max features: set 10000 as default for service
null for the collection min/maxscaledenom

backend_ogc_api_features

Metadataurl
- Klick auf das Metadata-Icon führt zu 404
klick_metadata

Style
- add legend

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesLoader.php
PhilVanB: [Fixed](https://github.com/mapbender/mapbender/pull/1831/changes/77763640846bc5a46924af55857c80ede2787bd8)

src/Mapbender/OgcApiFeaturesBundle/Entity/OgcApiFeaturesSource.php
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/5c0de26c96bed69c31035eb07da8716df21dd11f)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/53d00215edd7e13bbdcd3ac6759e1736001d6961)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
PhilVanB: This is necessary because if the IDs of a collection and the root layer happen to be the same, it will cause a messed up layer tree structure: the collection would then be displayed as the root layer. I added a comment.
I added [a comment.](https://github.com/mapbender/mapbender/pull/1831/changes/1f74c41e16a3cfcb4e1c9ba57adf8a2ca7befc46)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/999889935b46e38f30825e7eee208eb77607a9af)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesConfigGenerator.php
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/599624e52fa37763bd257a1f7c218d0e18baad48)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesInstanceFactory.php
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/2ed11759cc3d77c69b83bc1a4ffcfb358675a0ef)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesInstanceFactory.php
PhilVanB: [Fixed](https://github.com/mapbender/mapbender/pull/1831/changes/8b6f271628dad162529b791320e78cb1c7af19f6)

src/Mapbender/OgcApiFeaturesBundle/Resources/views/edit-instance.html.twig
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/757c740ea0d98b423919fe022c7c8654369fd8e4)

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesRenderer.php
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/a2ba751f1978ca378c64d2ede58ce10880d665f9)

src/Mapbender/OgcApiFeaturesBundle/Resources/views/edit-instance.html.twig
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1831/changes/7e750ee18a5fa36aa9854ea4e4518a9319d0b21f)

src/Mapbender/OgcApiFeaturesBundle/Resources/public/geosource.ogc_api_features.source.js
PhilVanB: [Fixed](https://github.com/mapbender/mapbender/pull/1831/changes/abcf438070ae619eb5f895dc77f0f6b78c527bd7)

src/Mapbender/OgcApiFeaturesBundle/Resources/public/geosource.ogc_api_features.source.js
PhilVanB: I added a [sorting](https://github.com/mapbender/mapbender/pull/1831/changes/bef2f2c44c54b9e4b4ccce64385e6159f1865737) for collections in layertree.

src/Mapbender/OgcApiFeaturesBundle/Resources/public/geosource.ogc_api_features.source.js
PhilVanB: Yes, WGS 84 is default, unless client explicitly requests other than that, e.g.:  GET /collections/{collectionId}/items?crs=http://www.opengis.net/def/crs/EPSG/0/25832

PhilVanB:
@astroidex Please check: 

- feature limit is now set to 1000 by default
- collection table: the input field for featureLimit now appears before the min-/max-Scale-fields
- collection table: the title field is wider now; min-/max-Scale and featureLimit-fields are smaller now

Phocacius (approved):

astroidex (approved):

PR #1832, opened 2026-03-16
Feature/ogc api datasource styling

frwg:
**Style Manager**

Central Mapbox style management (create, edit, copy, view, delete)

**OGC API Features — Styling**

Assign styles per layer with in-map picker and preview


src/Mapbender/ManagerBundle/Controller/RepositoryController.php
```diff
@@ -94,6 +96,11 @@ public function new(Request $request, string $sourceType): Response
                 $this->em->flush();
                 $this->em->commit();
 
+                if ($loader instanceof OgcApiFeaturesLoader) {
```
Phocacius: please don't import from a feature bundle in a core bundle, since the bundles can be deacivated individually, this might break. Instead, define an interface StyleableSource (or similar) in the core bundle and implement it in the APIFeaturesBundle.

src/Mapbender/ManagerBundle/Controller/RepositoryController.php
```diff
@@ -149,14 +156,21 @@ public function view($sourceId)
...
+        $viewData = array(
             'source' => $source,
             'applications' => $related,
             'title' => $dataSource->getLabel(true) . ' ' . $source->getTitle(),
             'grants' => $grants,
             // in backend, show all urls
             'secureUrls' => false,
-        ));
+        );
+        if ($source->getType() === 'ogc_api_features') {
```
Phocacius: this should use the interface as well then. The code within the if is specific to OGC API and therefore should live in the bundle.

src/Mapbender/ManagerBundle/Controller/RepositoryController.php
```diff
@@ -210,6 +224,15 @@ public function delete(Request $request, $sourceId)
             $affectedApplication->setUpdated($dtNow);
         }
 
+        // Delete associated styles before removing the source
+        $styles = $this->em->getRepository(Style::class)->findBy([
+            'sourceId' => $source->getId(),
+            'sourceType' => 'ogc_api',
```
Phocacius: IMO the check for source type can be removed, if in the feature we have other vector data sources, they will also be referenced by just the id.

src/Mapbender/ManagerBundle/Controller/RepositoryController.php
```diff
@@ -251,6 +274,11 @@ public function updateform(Request $request, $sourceId)
                 $this->em->flush();
                 $this->em->commit();
 
+                if ($loader instanceof OgcApiFeaturesLoader) {
```
Phocacius: see above

src/Mapbender/ManagerBundle/Controller/StyleController.php
```diff
@@ -0,0 +1,252 @@
...
+
+    #[ManagerRoute('/new', name: 'mapbender_manager_style_new', methods: ['GET', 'POST'])]
+    public function new(Request $request): Response
+    {
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_CREATE_STYLES);
+        $style = new Style();
+        $style->setSourceType('manual');
+
+        // OpenLayers default style as initial value
+        $defaultStyle = json_encode([
```
Phocacius: I'd extract this into a (protected) method, the default style is prone to be overridden is client projects.

src/Mapbender/ManagerBundle/Controller/StyleController.php
```diff
@@ -0,0 +1,252 @@
...
+    #[ManagerRoute('/{id}/edit', name: 'mapbender_manager_style_edit', methods: ['GET', 'POST'])]
+    public function edit(Request $request, int $id): Response
+    {
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_EDIT_STYLES);
+        $style = $this->em->getRepository(Style::class)->find($id);
+        if (!$style) {
+            throw $this->createNotFoundException();
+        }
+
+        if ($style->getSourceType() !== 'manual' && !in_array($style->getSourceType(), ['mapbox-json', 'sld'], true)) {
```
Phocacius: wieso ist manual hier extra aufgeführt und nicht Teil des in_array?

src/Mapbender/ManagerBundle/Controller/StyleController.php
```diff
@@ -0,0 +1,252 @@
...
+            return $this->redirectToRoute('mapbender_manager_style_view', ['id' => $id]);
+        }
+
+        if ($request->isMethod('POST')) {
+            $name = trim($request->request->get('name', ''));
```
Phocacius: Gibts einen Grund, die Validierung etc. hier manuell zu machen anstatt Symfony Forms zu nutzen?

src/Mapbender/ManagerBundle/Controller/StyleController.php
```diff
@@ -0,0 +1,252 @@
...
+    #[ManagerRoute('/{id}/delete', name: 'mapbender_manager_style_delete', methods: ['POST'])]
+    public function delete(Request $request, int $id): Response
+    {
+        $this->denyAccessUnlessGranted(ResourceDomainInstallation::ACTION_DELETE_STYLES);
+        if (!$this->isCsrfTokenValid('style_delete', $request->request->get('_token'))) {
+            throw new InvalidCsrfTokenException();
+        }
+        $style = $this->em->getRepository(Style::class)->find($id);
+        if ($style) {
+            if ($style->getSourceType() === 'ogc_api' && $style->getSourceId()) {
```
Phocacius: here too I'd remove the check for source type and just check if it has a source id for future compatibility

src/Mapbender/ManagerBundle/Controller/StyleController.php
```diff
@@ -0,0 +1,252 @@
...
+            $this->em->flush();
+            $this->addFlash('success', $this->trans->trans('mb.ogcapifeatures.admin.style.deleted'));
+        }
+
+        return $this->redirectToRoute('mapbender_manager_style_index');
+    }
+
+    private function extractVisualProps(Style $style): array
+    {
+        $defaults = [
```
Phocacius: this should reuse the defaults from line 70

src/Mapbender/ManagerBundle/Resources/public/css/style-editor.css
```diff
@@ -0,0 +1,173 @@
...
+    transition: all 0.15s ease;
+    cursor: pointer;
+}
+.style-editor-tabs .nav-tabs .nav-link:hover {
+    color: #495057;
+    background: #e9ecef;
+    border-color: #dee2e6 #dee2e6 transparent;
+}
+.style-editor-tabs .nav-tabs .nav-link.active {
+    color: #0d6efd;
```
Phocacius: use `var(--primary)` to get the mapbender default blue (which can also be overridden for client projects)

src/Mapbender/ManagerBundle/Resources/public/css/style-editor.css
```diff
@@ -0,0 +1,173 @@
...
+}
+#layer-accordion .accordion-button {
+    font-size: 13px;
+    padding: 8px 14px;
+    background: #f8f9fa;
+    font-weight: 600;
+}
+#layer-accordion .accordion-button:not(.collapsed) {
+    background: #e7f1ff;
+    color: #0d6efd;
```
Phocacius: use `var(--primary)` too

src/Mapbender/ManagerBundle/Resources/views/Style/edit.html.twig
```diff
@@ -0,0 +1,415 @@
...
+                    </fieldset>
+                </div>
+            </div>
+        </div>
+
+        <div class="mt-4 d-flex gap-2">
+            {% if not readonly|default(false) %}
+                <button type="submit" class="btn btn-primary"><i class="fas fa-save me-1"></i>{{ 'mb.ogcapifeatures.admin.style.editor.save' | trans }}</button>
+            {% endif %}
+            <a href="{{ path('mapbender_manager_style_index') }}" class="btn btn-secondary">{{ readonly|default(false) ? 'mb.ogcapifeatures.admin.style.editor.back' | trans : 'mb.ogcapifeatures.admin.style.editor.cancel' | trans }}</a>
```
Phocacius: The back button is always btn-light, btn-secondary is black on dark gray which is not readable

src/Mapbender/ManagerBundle/Resources/public/js/service-list.js
```diff
@@ -15,4 +15,47 @@ $(function() {
...
+                var $item = $(this);
+                if ($item.text().toUpperCase().indexOf(textVal.toUpperCase()) === -1) {
+                    $item.hide();
+                }
+            });
+        }
+    });
+
+    // re-apply type filter after text filter keyup
+    $(document).on('keyup', '.listFilterInput[data-filter-target="listFilterServices"]', function() {
```
Phocacius: Adding a second listener and applying the filter after a timeout seems very hacky and non-obvious when checking the filter code in components.js. I'd suggest merging the two listeners into one.

src/Mapbender/ManagerBundle/Resources/public/js/sld-converter.js
```diff
@@ -0,0 +1,237 @@
...
+ *   - TextSymbolizer (labels, halos, placement)
+ *   - RasterSymbolizer
+ *   - External graphic references (OnlineResource marks)
+ *   - Multiple Rule sets / SLD filters (only first rule used)
+ *   - Vendor-specific extensions (se:VendorOption etc.)
+ *   - Complex expressions or property-based styling
+ *
+ * Exposes window.Mapbender.SldConverter
+ */
+class SldConverter {
```
Phocacius: Did you review the GeoStyler library and their conversion functionality? Reducing external depencencies is basically a good thing, reinventing the wheel and having to maintain the functionality ourselves is a problem too though.

src/Mapbender/ManagerBundle/Resources/public/js/style-editor-layers.js
```diff
@@ -0,0 +1,609 @@
...
+        const ctx = c.getContext('2d');
+        const w = c.width, h = c.height;
+        ctx.clearRect(0, 0, w, h);
+
+        const p = layer.paint || {};
+        const lo = layer.layout || {};
+        const type = layer.type;
+
+        if (type === 'fill') {
+            const color = p['fill-color'] || '#cccccc';
```
Phocacius, line 460: where do the default values come from?

src/Mapbender/ManagerBundle/Resources/public/js/style-editor.js
```diff
@@ -0,0 +1,326 @@
...
+        this._bindFormValidation();
+    }
+
+    _bindFormValidation() {
+        const form = this.styleTextarea.closest('form');
+        if (!form) return;
+
+        form.addEventListener('submit', (e) => {
+            const raw = this.styleTextarea.value.trim();
+            if (!raw) return;
```
Phocacius: a return in the submit event listener will not prevent the form from submitting (only a return false (deprecated) or e.preventDefault() will do that)

src/Mapbender/ManagerBundle/Resources/public/js/style-editor.js
```diff
@@ -0,0 +1,326 @@
...
+
+    drawPreview() {
+        const json = this.getJsonFromTextarea();
+        if (this._isMultiLayerStyle(json)) return;
+        this._preview.drawAll(this.getVisualStyle());
+    }
+
+    updateRangePreviews() {
+        this.visualPane.querySelectorAll('input[type="range"]').forEach(input => {
+            const vp = input.closest('.mb-3').querySelector('.value-preview');
```
Phocacius: please don't use a style selector like `mb-3` for accessing an element in JS. Instead, use a dedicated class for it, that will not change when bootstrap changes its classes or the margin changes

src/Mapbender/ManagerBundle/Resources/views/macros/source.html.twig
```diff
@@ -3,7 +3,22 @@
...
+              {%- elseif type_name == 'WMTS' -%}
+                  <span class="badge bg-info text-white">OGC WMTS{% if source.version is defined and source.version %} {{ source.version }}{% endif %}</span>
+              {%- elseif type_name == 'TMS' -%}
+                  <span class="badge bg-warning text-white">TMS{% if source.version is defined and source.version %} {{ source.version }}{% endif %}</span>
+              {%- elseif type_name == 'vector_tiles' -%}
+                  <span class="badge bg-dark">Vector Tiles</span>
+              {%- elseif type_name == 'ogc_api_features' -%}
+                  <span class="badge bg-success">OGC API Features v1</span>
+              {%- else -%}
+                  <span class="badge bg-secondary">{{ source.type | source_label }}{% if source.version is defined and source.version %} {{ source.version }}{% endif %}</span>
```
Phocacius: This does not belong in the core bundle, mainly because this is not extendable by a custom source in a client project. Just call DataSource::getLabel for the label and either add a `getAccentColor` or similar to the DataSource abstract class for the background or "round robin" them assigning the colours as they occur.

src/Mapbender/ManagerBundle/Resources/views/Repository/index.html.twig
```diff
@@ -35,7 +35,23 @@
...
+                        <div class="text-end">
+                            <div class="content-box content-box-md">
+                                <div class="filter-group d-flex align-items-center justify-content-end gap-2">
+                                    <select id="filter-source-type" class="form-select d-inline-block w-auto" title="{{ 'mb.manager.admin.element.type' | trans }}">
+                                        <option value="">{{ 'mb.manager.source.filter_all_types' | trans }}</option>
+                                        <option value="WMS">OGC WMS</option>
+                                        <option value="WMTS,TMS">OGC WMTS / TMS</option>
+                                        <option value="vector_tiles">Vector Tiles</option>
+                                        <option value="ogc_api_features">OGC API Features</option>
+                                    </select>
```
Phocacius, line 46: Same here, these options should be dynamic. Use TypeDirectoryService::getSources to get a list of sources.

src/Mapbender/ManagerBundle/Resources/views/Style/edit.html.twig
```diff
@@ -0,0 +1,415 @@
...
+                    <a class="nav-link" id="tab-json" data-bs-toggle="tab" data-bs-target="#pane-json" role="tab">
+                        <i class="fas fa-code tab-icon"></i>{{ 'mb.ogcapifeatures.admin.style.editor.tab_json' | trans }}
+                    </a>
+                </li>
+                <li class="nav-item" role="presentation">
+                    <a class="nav-link" id="tab-import" data-bs-toggle="tab" data-bs-target="#pane-import" role="tab">
+                        <i class="fas fa-file-import tab-icon"></i>{{ 'mb.ogcapifeatures.admin.style.editor.tab_import' | trans }}
+                        <span id="import-tab-warning" class="d-none badge bg-danger ms-1" title="{{ 'mb.ogcapifeatures.admin.style.editor.import_overwrite_badge' | trans }}">!</span>
+                    </a>
+                </li>
```
Phocacius, line 44: I'd be in favour of removing the import tab in readonly mode, there's nothing you can do here anyways, so it's mainly confusing.

src/Mapbender/ManagerBundle/Resources/views/Style/edit.html.twig
```diff
@@ -0,0 +1,415 @@
...
+                <div class="tab-pane fade show active" id="pane-visual" role="tabpanel">
+
+                    {# ── Mapbox Layer Editor (shown for Mapbox documents) ── #}
+                    <div id="layers-editor" class="d-none">
+                        <div class="d-flex justify-content-between align-items-center mb-3">
+                            <span class="text-muted fw-bold text-uppercase" style="font-size:12px;letter-spacing:0.5px;">
+                                <i class="fas fa-layer-group me-1"></i>{{ 'mb.ogcapifeatures.admin.style.editor.mapbox_layers' | trans }}
+                            </span>
+                            <button type="button" class="btn btn-sm btn-outline-primary" id="btn-add-layer">
+                                <i class="fas fa-plus me-1"></i>{{ 'mb.ogcapifeatures.admin.style.editor.add_layer' | trans }}
```
Phocacius: Why is it only possible to add fill layers by using this button?

src/Mapbender/ManagerBundle/Resources/views/Style/edit.html.twig
```diff
@@ -0,0 +1,415 @@
...
+                    {# ── Flat Editor (shown for simple style objects) ── #}
+                    <div id="flat-editor">
+                    <fieldset>
+                        <legend><i class="fas fa-fill-drip me-1"></i>{{ 'mb.ogcapifeatures.admin.style.editor.fill' | trans }}</legend>
+                        <div class="row">
+                            <div class="col-sm-4">
+                                <div class="mb-3">
+                                    <label class="form-label fw-bold">{{ 'mb.ogcapifeatures.admin.style.editor.color' | trans }}</label>
+                                    <div class="input-group -js-colorpicker">
+                                        <input data-prop="fillColor" class="form-control" type="text" value="{{ _visual.fillColor ?? '#ff0000' }}">
```
Phocacius: for a fill-color with "stops", the editor only shows "[object Object]". Can be seen in the layer "Siedlung: Industrie und Gewerbe" in the Style "Daten des Liegenschaftskatasters in NRW - Nutzung"

src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesLoader.php
```diff
@@ -68,10 +71,167 @@ public function refreshSource(Source $source, mixed $formData): void
...
+                $entry['title'] = $titles[$key];
+            }
+            $result[] = $entry;
+        }
+        return $result;
+    }
+
+    private function discoverPropertyTitles(string $baseUrl, string $collectionId): array
+    {
+        $queryablesUrl = $baseUrl . '/collections/' . urlencode($collectionId) . '/queryables?f=json';
```
Phocacius, line 115: when there's queryables, is the lookup before (where fetching a single item) even necessary?

src/Mapbender/OgcApiFeaturesBundle/Entity/OgcApiFeaturesInstance.php
```diff
@@ -62,104 +77,94 @@ public function getSource(): Source
...
-        return $this;
     }
 
     public function getLayers()
     {
         return $this->layers;
     }
 
-    public function addLayer(OgcApiFeaturesInstanceLayer $layer): static
+    public function addLayer(OgcApiFeaturesInstanceLayer $layer): void
```
Phocacius: why disallow chaining here?

src/Mapbender/ManagerBundle/Resources/public/js/sld-converter.js
frwg: Should we follow the same architectural pattern for the GeoStyler integration as we did with ol-mapbox-style?

src/Mapbender/ManagerBundle/Resources/views/Repository/index.html.twig
```diff
@@ -35,7 +35,26 @@
                 </div>
                 {%- if items -%}
                     <div class="col-sm-7 col-xs-12 col-12">
-                    {% include "@MapbenderManager/fragments/filter-container.html.twig" with {filterTargetId: list_id} only %}
+                        <div>
+                            <div class="content-box content-box-md">
+                                <div class="filter-group d-flex align-items-center gap-2">
+                                    <div class="dropdown d-inline-block" id="filter-source-type-wrapper">
+                                        <select id="filter-source-type" class="hiddenDropdown" title="{{ 'mb.manager.admin.element.type' | trans }}">
+                                            <option value="">{{ 'mb.manager.source.filter_all_types' | trans }}</option>
```
Phocacius: filtering for the WMS type no longer works, for all other types it's fine

src/Mapbender/ManagerBundle/Resources/views/Repository/index.html.twig
```diff
@@ -35,7 +35,26 @@
                 </div>
                 {%- if items -%}
                     <div class="col-sm-7 col-xs-12 col-12">
-                    {% include "@MapbenderManager/fragments/filter-container.html.twig" with {filterTargetId: list_id} only %}
+                        <div>
+                            <div class="content-box content-box-md">
+                                <div class="filter-group d-flex align-items-center gap-2">
+                                    <div class="dropdown d-inline-block" id="filter-source-type-wrapper">
```
Phocacius: The box should be right-aligned: 

Image

src/Mapbender/OgcApiFeaturesBundle/Resources/public/geosource.ogc_api_features.source.js
```diff
@@ -19,17 +19,140 @@ class OgcApiSource extends Mapbender.Source {
...
+        }
+        // Mapbox style format — use ol-mapbox-style library (now bundled in ol.mapboxStyle)
+        if (styleDef.version && styleDef.layers && typeof ol !== 'undefined' && ol.mapboxStyle && ol.mapboxStyle.stylefunction) {
+            this._applyMapboxStyle(vectorLayer, styleDef, collectionId);
+            return;
+        }
+    }
+
+    _createSimpleOlStyle(s) {
+        const fillColor = this._hexToRgba(s.fillColor || '#ff0000', s.fillOpacity ?? 1);
```
Phocacius: Red as a fallback style seems to in-your-face to me. Maybe the default ol blue?

Phocacius (changes requested):

Phocacius (approved):

astroidex (approved):
Works fine

PR #1834, opened 2026-03-17
Add new parameter mapbender.search.default_regex for default search validation regex

Phocacius:
Follow-up to #1824: The default regex is now configurable using the new parameter `mapbender.search.default_regex` which default to `.*`, allowing all characters (as it was before). This default may be adjusted for the next major release, you can already use it to restrict inputs though. Also, the `pattern` parameter can be used per input field to override this value.


astroidex (approved):
Tested. Works fine.

astroidex (approved):
Tested. Works fine.

PR #1835, opened 2026-03-19
[ViewManager] Fix layerset generation

PhilVanB:
Fixes broken layersets after view loading
Solution: Also delete all layersets and then regenerate them correctly


Phocacius (approved):

astroidex (approved):
Works fine.

PR #1837, opened 2026-03-24
Fix layer sorting

PhilVanB:
- Fixes broken layer sorting after WMS (via WMS-Loader) or new view (via ViewManager) was loaded.
- See internal issue #11121
- Issue was caused by changes in [PR #1820](https://github.com/mapbender/mapbender/pull/1820)
- In this case, the solution of resetting `_sortableInitialized` back to `false` is fine. After all, it’s only called once. Previously, it was unnecessarily called 6–7 times or more.


Phocacius (approved):

PhilVanB:
@astroidex Please test again. The issues mentioned should now be fixed.

astroidex (approved):
works fine. Thank you.

PR #1839, opened 2026-04-14
Fix activating visible layers by name

Phocacius:


kpko:
Can confirm this fixes the bug we were having (as discussed with @astroidex) 👍

astroidex (approved):

PR #1844, opened 2026-05-05
Make sandbox iframe attribute in the feature info dialog configurable

Phocacius:
Using the parameter `mapbender.featureinfo.iframe_sandbox_params`. Default value is `allow-scripts allow-popups allow-popups-to-escape-sandbox allow-downloads`. For list of options, see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox)

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/config/mapbender.yaml
+++ b/src/Mapbender/CoreBundle/Resources/config/mapbender.yaml
@@ -35,5 +35,6 @@ parameters:
 
     mapbender.show_proxied_service_urls: false
 
+    # sandbox iframe attribute in the feature info dialog, see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox for options
     mapbender.featureinfo.iframe_sandbox_params: 'allow-scripts allow-popups allow-popups-to-escape-sandbox allow-downloads'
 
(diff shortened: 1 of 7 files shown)


astroidex (changes requested):
- variable works fine
- it does not limit to the possible parameters
- I can define allow-hello-world

or 
    mapbender.featureinfo.iframe_sandbox_params: 'allow-scripts allow-popups allow-popups-to-escape-sandbox allow-downloads allow-montag allow-popups" class="evel"'

src/Mapbender/CoreBundle/Resources/config/mapbender.yaml
```diff
@@ -35,3 +35,5 @@ parameters:
 
     mapbender.show_proxied_service_urls: false
 
+    mapbender.featureinfo.iframe_sandbox_params: 'allow-scripts allow-popups allow-popups-to-escape-sandbox allow-downloads'
```
astroidex, line 39: please add a short description and the link to the option 

similar to PR infos

Using the parameter mapbender.featureinfo.iframe_sandbox_params. Default value is allow-scripts allow-popups allow-popups-to-escape-sandbox allow-downloads. For list of options, see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox)

astroidex (changes requested):

Phocacius:
@astroidex please re-check.

astroidex (approved):
- fixed: injection of classes
- parameters.yaml with comment fixed

- still possible  allow-hello-world as definition is still possible but does not make trouble

Phocacius:
Additional properties do not make any harm since they are just ignored, and should the spec be updated in the future we don't have to adjust anything.

PR #1846, opened 2026-05-11
Feature/backdrop image v5

astroidex:
- new backdrop image is showing https://fr.wikipedia.org/wiki/M%C3%B4le_du_Raoulic
- thanks to @kartoffelfee for providing the image


Phocacius (changes requested):
Please also change the comment in https://github.com/mapbender/mapbender/blob/master/src/Mapbender/CoreBundle/Resources/config/mapbender.yaml#L19

Phocacius (approved):

PR #1850, opened 2026-05-12
[ViewManager] Show date in native format

Phocacius:
Display's date in native format in ViewManager column.

E.g. British:

grafik

German:

grafik

:warning: Requires additional composer packages. They will be automatically added after merging and applying composer update, for testing they need to be added manually though:

```bash
bin/composer require twig/intl-extra
bin/composer require twig/extra-bundle
```

see internal ticket 12348


astroidex (changes requested):
- I checked de/en/es. Works fine
- I also checked the print queue. Date is fine too.

astroidex (approved):

PR #1851, opened 2026-05-15
Map element validation

Phocacius:
Improves form validation for the map element. When saving the map element, the following is validated now:
- type checks for all number-based fields
- correct formatting for SRS (EPSG:<digits>)
- check that the SRS is defined in the mb_core_srs-Table
- check that at least one layerset is active


astroidex (changes requested):
- validation check works fine

But: Before the change it was possible to define an ALIAS for the EPSG Code description via |

It works like: EPSG:4326|ALIAS

This is not possible with the change anymore. It should be still possible,

Image

astroidex (approved):
works fine

PR #1852, opened 2026-05-19
Fix/debug batch print

frwg:
- Extracted extractBatchFrames() for robust frame-data decoding
- Removed redundant $configuration parameter threading; pass Element directly
- Queue action returns 204 No Content on form submit
- Fixed dumpPrint() call signature
- LayerRendererGeoJson: skip features outside render extent
- MbPrint.js: filter vector features outside print-frame extent


astroidex:
- the pastebin is not visible in sidepane if the sidepane is small
- suggestion: instead of text "Löschen" - place the big pastebin in the table header (will save a lot of space

table_print

Phocacius (approved):
Codewise it's fine, can be merged after addressing @astroidex 's comments.

Phocacius:
I'd suggest just leaving the column header for "Löschen" empty, the trash icon is standard anyway and does not need an additional label.

PR #1853, opened 2026-05-19
Fix/ogc api label display and source update

frwg:
- ${property} label templates in simple styles now resolve against feature properties at render time instead of being displayed literally
- Source reload updates existing layer rows in place by collectionId instead of creating new ones
- Deduplicates pre-existing layer rows on reload, keeping the lowest id
- Removes layer rows for collections no longer returned by the API
- Removes Style rows for dropped collections after each reload
- Prevents duplicate Style rows when reloading sources with pre-existing duplicate layer rows


astroidex (approved):
- labeling works fine
- rest texted via update

Phocacius (approved):

PR #1854, opened 2026-05-20
Feat/minimap buttons

JulianDietzel:
(see internal ticket 11915):

> Overview - Klötzchen verschönern - aber wie? [...] 

(English: "Prettify Overview - Button - but how? [...]")

## Changes
- The icons functionality to switch between an icon for the closed and open state was fixed (previously it did not change)
- The icon was made white and given more padding to match the scale slider above it on the screen (a hover effect was added to the background)
- The button now attaches to the overview map when open

**Changes not directly addressing the issue**:

- The `$radius`-SCSS variable was replaced by a new `radius-md`-CSS variable. Three additional radius classes were added for ease of use (and necessity for the buttons radii): `radius-sm`, `radius-lg` and `radius-full`.

## Screenshots
image

image

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/sass/libs/_variables.scss
+++ b/src/Mapbender/CoreBundle/Resources/public/sass/libs/_variables.scss
@@ -11,7 +11,6 @@ $panelBorderColor: #888;
 
 //-----------------------------------------------------------------other
 $space: 20px;
-$radius: 3px;
 $popupModalWidth: 600px;
 $desktopBreakpointWidth: 1200px;
 
--- a/src/Mapbender/CoreBundle/Resources/public/sass/libs/_variables.scss
+++ b/src/Mapbender/CoreBundle/Resources/public/sass/libs/_variables.scss
@@ -121,4 +121,15 @@ $ciColor2: #4899ba; // Doesn't do anything
     --toolbar-text-color: var(--text-color);
     --toolbar-button-active-background-color: var(--primary);
     --toolbar-button-active-background-hover-color: var(--primary-dark);
+
+    // Buttons
+    --button-white-foreground: var(--text-color);
+    --button-white-background: var(--background-color);
+    --button-white-background-hover: #f7f7fa;
+
+    // Radii
+    --radius-sm: 2px;
+    --radius-md: 3px;
+    --radius-lg: 5px;
+    --radius-full: calc(infinity * 1px);
 }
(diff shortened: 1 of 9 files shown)


astroidex (changes requested):
- looks nice 
- Suggestion use far fa-map instead
- box size on closed overview state is fine
 

Image
 
 
- arrow (>) and white square on open stage look a bit big - please check with Mobile mode

Image

src/Mapbender/CoreBundle/Resources/public/sass/libs/_variables.scss
```diff
@@ -121,4 +120,15 @@ $ciColor2: #4899ba; // Doesn't do anything
...
+    // Buttons
+    --button-white-foreground: var(--text-color);
+    --button-white-background: var(--background-color);
+    --button-white-background-hover: #f7f7fa;
+
+    // Radii
+    --radius-sm: 2px;
+    --radius-md: 3px;
+    --radius-lg: 5px;
+    --radius-full: calc(infinity * 1px);
```
Phocacius: I'd call it `radius-circle` or similar to make it clearer what it does

JulianDietzel:
Thanks for the feedback.

> Suggestion use far fa-map instead

I tried this, that's what I wanted to use. But on my machine I couldn't the the regular font awesome icons to be displayed corretly. This was reproducable on @Phocacius machine as well. The only way to get it to work for me was to delete the `.woff2`-file of font awesome, falling back on the `.tiff`-file. After some probing around I deemed this to be a separate issue.
(For some reason it just started to work on @Phocacius machine again. Might need some investigation)

> arrow (>) and white square on open stage look a bit big

I tried viewing the Overview map in mobile mode but in my demo project it got hidden on mobile which made me (incorrectly) believe the overview just can't be shown on mobile. Now that I see it I am a bit torn. I agree that it looks a little big but then again it is especially on mobile where buttons shouldn't be too small for use on touch screens. Perhapse we could try making it a little slimmer while maintaining the height?

The earliest I will be able to address this will be in two weeks. So feel free to fix this in the meantime. Or leave it here and I'll deal with it in two weeks.

Phocacius:
I slighly decreased the icon size and also added a check if the overview is on the left or right to change the icon direction.

The problem with the icon was not the map (that does exist in both far and fas) but the chevron, which only exists is fas. I added a toggle to switch between fas and far.

astroidex (approved):

PR #1855, opened 2026-05-20
Do not show "abort" error message

Phocacius:
When e.g. quickly swiping through digitizer lists, the previous requests get cancelled. This should an "abort" error message to the user. However,  aborting the request is not in fact an error but intended behaviour. This PR removes these error displays.


astroidex (approved):
- I tested with 5000 points. Works fine.

PR #1856, opened 2026-05-20
[Forgot password] make check for email case-insensitive

Phocacius:


astroidex (changes requested):
- I can enter whatever I want and get the message that the email was send

I tried
- my email that is registered in Mapbender
- my email with upper/lower case
- a fake email
- I tried my user name
- I tried a fake user name
- awĺways get the message "Eine E-Mail mit Ihrem zurückgesetzten Code und Link ist Ihnen zu Ihrer bereitgestellten Adresse gesendet worden."
 
I do not have a mailer defined - older versions with no mailer defined I would get an error message

Phocacius:
> I can enter whatever I want and get the message that the email was send

That is intended behaviour so intruders can't brute-force/check if a user is registered. An email is only actually sent if a user with that mail exists in the database.

astroidex (approved):
works fine when mailer is defined. I receive the mail and can reset the password

PR #1858, opened 2026-06-09
Fix GpsPosition: Initialize stack array to fix geolocation error

swinkelmann:
Initialize this.stack, this.observer, and this.firstPosition in the _setup() 
method to prevent "Cannot read properties of undefined" errors when 
_handleGeolocationPosition() is called.


astroidex (approved):
works fine

Phocacius (approved):

PR #1859, opened 2026-06-09
Fix/poi textarea

swinkelmann:
The _sendPoi() method was reading the textarea value from the original
element container (this.$element) instead of from the active popup
where the user actually enters their input. This caused the method to
always return the initial placeholder text, ignoring any user input.

Also removes the “Close” button from the POI element's popup


astroidex (approved):
works fine

Phocacius (approved):

PR #1863, opened 2026-06-16
Sidebar improvements

Phocacius:
Fixes the following sidepane-related issues:

- Inline elements (HTML or Simple Search) were always at the top, regardless of their order in the sidepane
-  You could not enlarge the view by dragging with the mouse if an element is expanded in List mode
-  If the sidepane is on the right, the Close button overlapped the other buttons
- Sidepane Mode Button: Show icons as well
- When moving the slider on the WMS, the width of the sidepane increased
- Padding in the Sidepane modes "Buttons" und "unstyled" was missing
- on mobile, sidepane could not be closed when the initial width exceeded the display width

see internal tickets 13378 and 13383

### DIFF
--- a/src/Mapbender/CoreBundle/Element/ApplicationSwitcher.php
+++ b/src/Mapbender/CoreBundle/Element/ApplicationSwitcher.php
@@ -104,11 +104,11 @@ class ApplicationSwitcher extends AbstractElementService implements ConfigMigrat
                 if (empty($appConfig['url'])) {
                     $appConfig['url'] = $this->router->generate('mapbender_core_application_application', ['slug' => $slug]);
                 }
-                if (empty($appConfig['imgUrl'])) {
-                    $appConfig['imgUrl'] = false;
+                if (empty($appConfig['img_url'])) {
+                    $appConfig['img_url'] = false;
                     $imgPath = '/uploads/' . $slug . '/' . $application->getScreenshot();
                     if (@\is_file($this->rootDir . '/public' . $imgPath)) {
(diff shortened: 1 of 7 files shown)


astroidex (changes requested):
Fixes the following sidepane-related issues: work fine

- [x] Inline elements (HTML or Simple Search) were always at the top, regardless of their order in the sidepane
- [x] You could not enlarge the view by dragging with the mouse if an element is expanded in List mode
- [x] If the sidepane is on the right, the Close button overlapped the other buttons
- [x] Sidepane Mode Button: Show icons as well
- [x] When moving the slider on the WMS, the width of the sidepane increased
- [x] Padding in the Sidepane modes "Buttons" und "unstyled" was missing
- [x] on mobile, sidepane could not be closed when the initial width exceeded the display width

noticed behaviour
- sidepane x atthe top : please delete backgroundcolor - on mouseover on the heading the background behind x is visible
- akkordeon and list: sidepane mit inline html with lot of content: scrolling should be possible if there is a lot of content 
  - can not open elements in akkordeon mode with big  html inline 
  - can not see elements at the bottom in llist mode

Image

Phocacius:
Scrolling was implemented for list mode. For accordion mode, it's not much more complicated, since the space is equally distributed between the titles, with the remaining space dedicated to the active element. If the whole element scrolls, it's not obvious how big the active element is supposed to grow. This is not a regression though, the whole point of the list mode was to mitigate the problem of too much content in the sidepane, I'd say if vertical space is a problem, people should just use the list mode.

PR #1864, opened 2026-06-16
Interactive help improvements

Phocacius:
- Make "do not show again" button toggleable by clicking on text
- Wait for finished transitions before updating position to fix positions of elements pointing to sidebar elements in list mode  

Code refactor:
- convert sidepane.js into it's own class registered as Mapbender.sidePane with a public API like `setOpen` and `openElementById` for central access


astroidex (changes requested):
it can happen, that the sidepane is open (right side) and is hiding the overview element (same for scalebar, coordinate utility..). Then the help is on the blank sidepane. 

But if it is too complicated to close the sidepane at that moment it is ok to leave it like that

PhilVanB (approved):
Major code improvements, good job!

PR #1865, opened 2026-06-17
Refactor/scss to css vars

frwg:
- adjust layout of metadata header 
- change css class of gpsElement
- allow change of font settings via css vars
- see documentation at https://github.com/mapbender/mapbender/blob/develop/docs/design/css_custom_properties.md


astroidex:
@frwg please provide a sample how to modify the styles via variables now: blue would be nice

Phocacius (approved):

PR #1867, opened 2026-06-17
Enable Style definition in yaml file

frwg:
To verify this implementation locally, please follow these steps:

Navigate to the styles configuration directory: /application/config/styles.

Create a new test configuration file named mapbender_styles.yaml.

Add a sample configuration similar to this:

```
parameters:
    styles:
        demo_red_polygon:
            name: "Demo Red Polygon (YAML)"
            style:
                fillColor: "#ff0000"
                fillOpacity: 0.3
                strokeColor: "#990000"
                strokeOpacity: 1
                strokeWidth: 2
                pointRadius: 6
```


astroidex (commented):
For discussion 
- Does it make sense to separate the styles from the applications?
- till now everything is within an application definition

frwg:
@astroidex  I can see your issue, but the problem is, that styles are defined totally independent of applications. What we could do instead is define styles directly where they are actually used - so far only within the definition of OGC API Feature Sources within the layersets at every single layer, but then there were no way to re-use them.

astroidex (changes requested):
- [x] Works fine to define new styles via yaml defnition
- [x] Styles show up in the styles section and can be copied
- [ ] Styles can not be used directly in a service instance - is that ok? Just a question. But would be a nice way to prepare a collection of styles 

```
parameters:
    styles:
        red:
            name: "Red (YAML)"
            style:
                fillColor: "#ff0000"
                fillOpacity: 0.3
                strokeColor: "#990000"
                strokeOpacity: 1
                strokeWidth: 2
                pointRadius: 8
        blue:
            name: "Blue (YAML)"
            style:
                fillColor: "#00ff00"
                fillOpacity: 0.3
                strokeColor: "#00ff00"
                strokeOpacity: 1
                strokeWidth: 2
                pointRadius: 8
        yellow:
            name: "Yellow (YAML)"
            style:
                fillColor: "#ffff00"
                fillOpacity: 0.3
                strokeColor: "#ffff00"
                strokeOpacity: 1
                strokeWidth: 2
                pointRadius: 8

```

Phocacius (changes requested):
- To make the definition via YAML useful, IMO there should be a way to directly use it in a YAML-defined application (for sample definition, see https://github.com/mapbender/mapbender/pull/1831, maybe add a property `style` refering to the name of the style)
- In the style selection dialog, the YAML styles cannot be used directly. They are only shown in the styles submenu

Image

PR #1869, opened 2026-06-18
Fix toolbar button groups

swinkelmann:
Buttons are properly deactivated again when they are in a group

internal ticket: #13385


astroidex (approved):
works as expected

PhilVanB (approved):

Phocacius (approved):

PR #1870, opened 2026-06-18
Fix Layertree: Ignore not allowed layers on shift click

swinkelmann:
If the user is not permitted to deactivate an active layer, this is now taken into account when Shift+clicking on the root layer.

internal ticket: #13381


astroidex (approved):
works as expected

Phocacius (approved):

PR #1871, opened 2026-06-18
Fix/adjust ogcapi style

frwg:
- Improved Opacity Handling: Refactored the opacity logic for more consistent rendering.
- Enhanced Mapbox Variable Calculation: Fixed and optimized the calculation of variables derived from mathematical expressions within Mapbox styles.
- Added Required Field Validation: Required input fields now visually indicate missing values with a light red highlight.
- Supported Filtered Mapbox Styles: Enabled the display and rendering of features originating from filtered Mapbox styles.
- Adds/improves font styling options in the StyleEditor
- Popover on features adjust its size dynamically
- Optional HTML Templates for Popover and FeatureInfo
- Popover properties are sortable now
- Popover properties can be given a label that is used instead of the property name when given

### DIFF
--- a/src/Mapbender/ManagerBundle/Controller/StyleController.php
+++ b/src/Mapbender/ManagerBundle/Controller/StyleController.php
@@ -214,6 +214,7 @@ class StyleController extends ApplicationControllerBase
             'fontFamily' => 'Arial, Helvetica, sans-serif',
             'fontSize' => 11,
             'fontWeight' => 'regular',
+            'fontStyle' => 'normal',
             'fontColor' => '#333333',
             'fontOpacity' => 1,
         ];
--- a/src/Mapbender/ManagerBundle/Resources/public/css/style-editor.css
+++ b/src/Mapbender/ManagerBundle/Resources/public/css/style-editor.css
@@ -84,6 +84,13 @@
     margin-bottom: 6px;
 }
 
+/* ── Empty field validation ── */
+input.field-empty,
+input.field-empty:focus {
+    background-color: #fff0f0 !important;
+    border-color: #f5a0a0 !important;
+}
+
 /* ── Layer Accordion ── */
 #layers-editor .layers-header {
     display: flex;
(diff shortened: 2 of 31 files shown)


astroidex (changes requested):
- [ ] Improved Opacity Handling: Refactored the opacity logic for more consistent rendering.
**Feedback: works in the map**
**does not work in the preview Style editor

Image

Not sure how to test
- [ ] Enhanced Mapbox Variable Calculation: Fixed and optimized the calculation of variables derived from mathematical expressions within Mapbox styles.
- [ ] Added Required Field Validation: Required input fields now visually indicate missing values with a light red highlight. 

- [x ] Supported Filtered Mapbox Styles: Enabled the display and rendering of features originating from filtered Mapbox styles.
- [x ] Adds/improves font styling options in the StyleEditor

astroidex (changes requested):
When I try to import the mapbender_user application from yaml I get the error

Image

Impossible to access an attribute ("collectionId") on a null variable in @MapbenderOgcApiFeatures/edit-instance.html.twig at line 87.

frwg:
@astroidex 

> * [ ]  Improved Opacity Handling: Refactored the opacity logic for more consistent rendering.
>   **Feedback: works in the map**
>   **does not work in the preview Style editor

Hmm, this works pretty well in my application locally? Could you check again?

Phocacius:
@frwg opacity 0 in the style preview does not work for me either. Instead of transparent, it jumps to completely opaque again when going from 0.01 to 0 (both for fill and stroke)

Phocacius (changes requested):
- The popover HTML template does not work at all. If I click on the "HTML-Vorlage", there still only is the property list visible. When saving and opening the popup again, nothing is selected.

grafik

- Popover properties are sortable now: Does not work for me, when dragging in the property list, I only sort the whole collections row

- Where do all these adjustments come from and why are they necessary? The internal tickets 13390 and 13393 are only small parts of the changes in here. Please split the PRs accordingly and reference the corresponding tickets in the future.

Phocacius:
@frwg @astroidex Can this be closed after #1880  and #1881 ?

PR #1874, opened 2026-06-25
Replace checkbox with selectbox for menu type

PhilVanB:
This PR replaces the checkbox that generated a menu for the toolbar-buttons with a selectbox with the options _no menu_, _Show menu on desktop_, _Show menu on mobile_ and _Show menu on desktop+mobile_.

See internal #13389

(Outdated) selectbox-values for YAML-Config:

```
no_menu
menu_desktop
menu_mobile
menu_mobile_desktop
```

Update after #1893: New values for YAML-Config:

```
no
yes
only_desktop
only_mobile
```

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/mapbender.responsive-menu.js
+++ b/src/Mapbender/CoreBundle/Resources/public/mapbender.responsive-menu.js
@@ -1,0 +1,141 @@
+!(function($) {
+	const DESKTOP_BREAKPOINT = 1200;
+	const DROPDOWN_OFFSET = 3;
+
+	function getMenuItemMap(toolbar) {
+		const serialized = toolbar.dataset.menuItemMap || '[]';
+
+		if (toolbar._responsiveMenuMapSource !== serialized) {
+			toolbar._responsiveMenuMapSource = serialized;
+			try {
+				toolbar._responsiveMenuMap = JSON.parse(serialized);
+			} catch (error) {
+				toolbar._responsiveMenuMap = [];
+			}
+		}
+		return toolbar._responsiveMenuMap || [];
+	}
+
+	function isInlineMode(toolbar, screenWidth) {
+		if (toolbar.classList.contains('has-menu-mobile')) {
+			return screenWidth >= DESKTOP_BREAKPOINT;
+		}
+		if (toolbar.classList.contains('has-menu-desktop')) {
+			return screenWidth < DESKTOP_BREAKPOINT;
+		}
+		return false;
+	}
+
+	function resetMenuWrapperState(menuWrapper) {
+		const button = menuWrapper.querySelector('button');
+
+		menuWrapper.classList.remove('open');
+		if (!button) {
+			return;
+		}
+
+		button.classList.remove('active');
+		const icon = button.querySelector('i');
+		if (icon) {
+			icon.classList.add('fa-bars');
+			icon.classList.remove('fa-xmark');
+		}
+	}
+
+	function updateDropupLayout(menuWrapper) {
+		if (!menuWrapper || !menuWrapper.classList.contains('dropup')) {
+			return;
+		}
+
+		if (!menuWrapper.classList.contains('open')) {
+			menuWrapper.style.removeProperty('--dropdown-bottom');
+			menuWrapper.style.removeProperty('--dropdown-menu-max-height');
+			return;
+		}
+
+		const toolbar = menuWrapper.closest('.toolBar');
+		if (!toolbar) {
+			return;
+		}
+
+		const toolbarRect = toolbar.getBoundingClientRect();
+		const topToolbar = document.querySelector('.toolBar.top');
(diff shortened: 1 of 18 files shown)


src/Mapbender/FrameworkBundle/Component/Renderer/ApplicationMarkupRenderer.php
```diff
@@ -152,6 +152,29 @@ protected function getRegionTemplateVars(Application $application, $regionName,
...
+            $buttonMenu = $props['generate_button_menu'];
+            switch ($buttonMenu) {
+                case 'no_menu':
+                    $props['generate_button_menu'] = false;
+                    break;
+                case 'menu_desktop':
+                    $props['generate_button_menu'] = !(in_array($screenType, ['mobile', 'all']));
+                    break;
+                case 'menu_mobile':
+                    $props['generate_button_menu'] = !(in_array($screenType, ['desktop', 'all']));
```
Phocacius: The menu should be shown (or not) depending on the actual display resolution of the client that accesses the application (responsively). Right now, the menu still basically works as a fixed checkbox.

Phocacius (changes requested):

astroidex (approved):
tested. Works fine.

src/Mapbender/CoreBundle/Resources/public/mapbender.responsive-menu.js
```diff
@@ -0,0 +1,141 @@
+!(function($) {
```
Phocacius, line 1: This could be converted to a class (e.g. Mapbender.ResponsiveMenu). That way, it is possible to extend parts of the functionality in client apps (e.g. if the breakpoint should be different)

src/Mapbender/CoreBundle/Resources/views/Template/region/toolbar.html.twig
```diff
@@ -16,17 +28,49 @@
...
     {%- endif -%}
     {% block item_list %}
-    <ul class="{{ alignment_class | default('itemsRight') }}">
+    <ul class="{{ items_list_class }}" data-toolbar-inline-list>
         {% block item_list_body %}
             {{- inline_items | raw -}}
         {% endblock %}
     </ul>
     {% endblock item_list %}
+    {%- if menu_mode != 'no_menu' and elements is defined and elements|length -%}
```
Phocacius, line 37: Not sure if this is a regression or if it always was that way, but IMO the labels should always be shown in menu mode, independant of the "show labels" setting

Image

astroidex: it has been like that before. But I agree it would be good to show the label in menu mode.

src/Mapbender/CoreBundle/Resources/public/mapbender.responsive-menu.js
PhilVanB: [Done](https://github.com/mapbender/mapbender/pull/1874/changes/bbcc739aa6a25bac71c2bab427d6b2a2a3852ccc)

src/Mapbender/CoreBundle/Resources/views/Template/region/toolbar.html.twig
PhilVanB: Now labels are always displayed when responsive menu is in mobile mode. Take a look [here](https://github.com/mapbender/mapbender/pull/1874/changes/290b603e29bcdcc08e01477456d4a7d2d5b122e2).

Phocacius: If in responsive mode (e.g. mobile only) the labels are now also displayed when in expanded mode, that's also not a desired behaviour.

astroidex (changes requested):
- tested in header and footer. Works as expected. 

But when I use in yaml definition:

`generate_button_menu: huhu`

It will always show the menu

- default should be generate_button_menu: false or generate_button_menu: no_menu
- when the user defines QUATSCH. The menu should not be shown

PhilVanB:
> * tested in header and footer. Works as expected.
> 
> 
> But when I use in yaml definition:
> 
> `generate_button_menu: huhu`
> 
> It will always show the menu
> 
>     * default should be generate_button_menu: false or generate_button_menu: no_menu
> 
>     * when the user defines QUATSCH. The menu should not be shown

[Fixed](https://github.com/mapbender/mapbender/pull/1874/changes/5f70c30106ef4096761e0456243e0505b5abea7b)

src/Mapbender/CoreBundle/Resources/views/Template/region/toolbar.html.twig
PhilVanB: [Fixed](https://github.com/mapbender/mapbender/pull/1874/changes/a39808d18817d1eb45d5bf7347ca975557972e8c)

astroidex (approved):
Tested. Works fine.

astroidex (approved):

Phocacius (approved):

PR #1877, opened 2026-06-30
Restore popup structure and hide footer buttons when minimized on mobile

swinkelmann:
Reverted the popup HTML structure to its original state and implemented an alternative solution to correctly hide the footer buttons when the popup is minimized on mobile devices.

### DIFF
--- a/src/Mapbender/CoreBundle/Resources/public/sass/modules/_popup.scss
+++ b/src/Mapbender/CoreBundle/Resources/public/sass/modules/_popup.scss
@@ -227,6 +227,13 @@ body.popup-resizing {
     }
   }
 
+  .popup.popup-minimized {
+    .popup-body,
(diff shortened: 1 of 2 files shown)


astroidex (commented):
Works fine for 
- [x] searchrouter
- [x] meetingPoint
- [x] WMSloader
- [x] featureInfo

But does not work for

- [ ] printClient
- [ ] batchprint
- [ ] imageExport

astroidex (approved):
@swinkelmann said it should be ok for the missing elements. So I approve

swinkelmann:
The buttons in printClient and imageExport are not buttons in the footer of the popup. They are submit buttons on the print form itself. 
However, these are also hidden when the popup is minimised.

Phocacius (approved):

PR #1880, opened 2026-07-07
Fix/style opacity zero

frwg:
Fixed opacity value handling so an explicit 0 is preserved (instead of falling back to 1), ensuring transparent styles render correctly and consistently across previews.
internal Ticket 13390


Phocacius (approved):

astroidex (commented):
Tested and approved

PR #1881, opened 2026-07-07
Fix/style font weight style

frwg:
Style Editor: fixes font weight/style handling so bold, italic, and bold+italic render correctly.
Custom style labels now render consistently between editor preview and runtime rendering.


frwg:
Accidentally incorrect base branch selected

PR #1883, opened 2026-07-08
[ApplicationSwitcher] Fix various issues

PhilVanB:
Fixes the following issues with the application switcher:

- when a WMS is added the application now remains and is not opened in a new browser-tab
- issues with the attributes `data-mb-layer-merge` and `data-mb-wms-merge` are now fixed and work properly
- when a WMS is added the bounding box of the map now remains the same an is not changed to the start extent


Phocacius (approved):

astroidex (approved):
works as expected

PR #1886, opened 2026-07-09
Hide loading spinner when a layer is deactivated in the layer tree

Phocacius:
When a source is deactivated in the layertree while it was still loading, the loading spinner was shown indefinetly. This PR fixes that and instantly hides the loading spinner when a source was deactivated.


astroidex (approved):
Tested. Works fine.

PR #1891, opened 2026-07-14
Fix print order when ogc api features sources are involved

Phocacius:
OGC API Features was always printed on top, even when in the layer tree is was not on top.

Reason: The Vector Layers were manually added after all the other layers in the MbImageExport::_collectGeometryLayers method.

Solution: Add a flag `geojson-mblayer` for vector layers that are actual sources and ignore them in MbImageExport::_collectGeometryLayers. Add a printConfig for OGC API Features so the layers are exported at the right position. Extract dumping of vector layer data to geojson to Mapbender.Model to avoid code duplication.

see internal ticket 13630


astroidex (approved):
Looks like ordering looks good.

But found problems with 
- Styles with text
- Styles created with Mapbender
-  see internal  13637

Phocacius:
The print style problems are unrelated to the ordering fixed here, will be addressed in another PR.

PR #1892, opened 2026-07-15
Fix "flickering" of non-inline html items in content area

Phocacius:
see internal ticket 13564


astroidex (approved):
Tested. Works fine.

PR #1894, opened 2026-07-16
Increase backend popup size and make it configurable

Phocacius:
- Increase backend popup size and make it configurable

new CSS Variables:

```css
    --backend-popup-max-width: 800px;
    // allows popups to grow bigger than --backend-popup-max-width on large screens
    --backend-popup-max-width-override: 65vw;
```

- When restricting frontend popup placement after movement, also restrict size


astroidex (changes requested):
Looks a bit strange

Width too high

Image

astroidex (approved):

PR #1898, opened 2026-07-20
Add new option mapbender.application.sortorder for sorting of application list

Phocacius:
To avoid having the demo application always on top of the (usually more useful) db apps, the default sorting of the application list was changed to having the db applications shown first. However, the parameter `mapbender.application.sortorder` was added to control this behaviour:

| Parameter Value | Description                                                                                                                                                                                                                                                                              |
|-----------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| db-first        | New default: Db applications are shown first, then yaml applications. Internally sorted by title                                                                                                                                                                                         |
| yaml-first      | Previous default: Yaml applications are shown first, then db applications. Internally sorted by title                                                                                                                                                                                    |
| db-only         | Only db applications are shown in the applications list, yaml applications are hidden. :warning: yaml applications are still accessible by calling `/application/<slug>`, to completely remove yaml applications, delete the from the `config/applications` directory in the filesystem. |
| yaml-only       | Only yaml applications are shown in the applications list, db applications are hidden. :warning: db applications are still accessible by calling `/application/<slug>`, to completely remove db applications, temporarily change the parameter value back and delete the applications    |
| title           | Applications are sorted by title, independant of origin                                                                                                                                                                                                                                  |
| date            | Applications are sorted by most recent update, independant of origin                                                                                                                                                                                                                     |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Other:
 * [Develop] Add ExternalGraphicUtil (previously part of the digitizer module) ([PR#1897](https://github.com/mapbender/mapbender/pull/1897))
 * In applications tab in source infos, use same symbology for public/not public as elsewhere in Mapbender ([#PR1812](https://github.com/mapbender/mapbender/pull/1812)) 
 * Changed default login-backdrop image ([PR#1845](https://github.com/mapbender/mapbender/pull/1845))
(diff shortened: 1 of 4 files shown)


astroidex (approved):
Tested. Works fine.

PR #1900, opened 2026-07-22
Fix sidepane was still shown when all items where hidden by responsive options in unstyled mode

Phocacius:
fixes internal ticket 13682


PhilVanB (approved):
Works as expected

PR #1902, opened 2026-08-03
Fix mapbender:wms:assign command

PhilVanB:
Fixes the `mapbender:wms:assign` command: Now passes the correct parameter

see internal issue #13820


astroidex (approved):

astroidex (commented):
Tested. works fine.

PR #1909, opened 2026-08-17
Add css variable for toolbar menu text color

Phocacius:
Makes the default colour for the toolbar menu `--toolbar-text-color`. It is possible to override the text color for the menu only be the new variable `toolbar-menu-text-color`.

Example:

```css
:root {
  --toolbar-text-color: orange;
  --toolbar-menu-text-color: green;
}
```


astroidex (approved):
Tested. Works fine

PR #1913, opened 2026-08-19
Make textarea in manager (e.g. for search) resizable, add border to code mirror textareas

Phocacius:


astroidex (approved):
works fione. Would be nice for digitizer too

Phocacius:
The digitizer window grows automatically anyways, since it uses code mirror. For the digitzer (and e.g. InteactiveHelp) the improvement is the border around the textarea.

PR #1914, opened 2026-08-19
Fix POI function calls to use lon/lat parameters

swinkelmann:
Fixes a bug that was accidentally caused by code completion in the IDE. Sorry!


astroidex (approved):
tested. works fine.

PR #1915, opened 2026-08-19
Fix OgcApiFeaturesLayerSource had no id and was therefore ignored when exporting or duplicating a YAML application

Phocacius:
for testing: duplicate the new mapbender_user.yaml (from https://github.com/mapbender/mapbender-starter/pull/175)

### DIFF
--- a/src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesInstanceFactory.php
+++ b/src/Mapbender/OgcApiFeaturesBundle/Component/OgcApiFeaturesInstanceFactory.php
@@ -107,6 +107,7 @@ class OgcApiFeaturesInstanceFactory extends SourceInstanceFactory
                 continue;
             }
             $layerSource = new OgcApiFeaturesLayerSource();
+            $layerSource->setId($layer['collectionId']);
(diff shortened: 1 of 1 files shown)


astroidex (changes requested):
- Question: for Info all attributes are active, for hover you have to activate them. Shall we deactivate them for Info too, so the user has toactivate the ones needed?

FeatureInfo
- html: Suggestion: Result should not contain the Collection-Title. If someone wants a title. This can be defined & styled in the html

For discussion. The rest works fine

Phocacius:
> Question: for Info all attributes are active, for hover you have to activate them. Shall we deactivate them for Info too, so the user has toactivate the ones needed?

That was intentional: A tooltip is only a preview, it should only show the most relevant info (as selected by the user). FeatureInfo shows the comprehensive details of a feature, and QGIS also shows all attributes by default (when not using the maptip feature but the default list style). I may be convinced otherwise :)

As this PR is for the bugfix that a YAML application was not correctly imported into the db, I'll merge this PR anyways, and copy this discussion to https://github.com/mapbender/mapbender-starter/pull/175

PR #1916, opened 2026-08-19
Do not show two delete popups when deleting a source

Phocacius:


astroidex (changes requested):
in sidepane a lot of options are visible that do not have an effect.

Image

PR #1917, opened 2026-08-20
Do not show title in feature info custom html

Phocacius:
it can be added manually to the custom html if it is desired. It is still shown for YAML and property lists mode.

Also improved the mechanism to detect whether there were replacements made: The StyleUtil now has tracks how many variables were found and how many of them were replaced. If there was no replacement made, the featureinfo is not shown, unless the template is static (i.e. no variables exist)


astroidex (approved):
Tested works fine

PR #1918, opened 2026-08-20
Allow print preview for all data types 

Phocacius:
This works by injecting html into empty popup window.

Supports both OGC API Features and Vector Tiles.


astroidex (approved):
Tested. Works fine

PR #1919, opened 2026-08-20
Fix loading of saved OGC API Features layers in the ViewManager

swinkelmann:
Saved layers were nested twice when a view was loaded, which prevented the service from loading correctly. 
The ViewManager now removes the additional layer wrapper before recreating the source.


astroidex (approved):
Handling of OGC API features now works fine

PR #1920, opened 2026-08-20
Fix metadata for ogc api features

Phocacius:


astroidex (changes requested):
Metadata for layer 8collection) works fine

But
- error: wehn you try to get the metadata for the whole service (root layer) - **message: Invalid instance layer id**

astroidex (approved):

PR #1921, opened 2026-08-20
In MbButton use existing checkAutoOpen method to check for initial state; override in MbInteractiveHelp to account for "don't show again"

Phocacius:


astroidex (approved):
tested. works fine

PR #1922, opened 2026-08-20
Fix logo sizing and spacing

swinkelmann:
Removes toolbar padding constraints from the application logo while
preserving the application name spacing, allowing custom logos to be
displayed at their configured size. 

configuration example:
.logoContainer .logo {
  height: 28px;
  padding: 5px
}

internal ticket: 13879


astroidex (approved):
works fine. Has still the padding of the toolbar but is ok for now.

swinkelmann:
You can also remove the padding of the toolbar by configuration:
.logoContainer {
    margin: calc(var(--space) / -4) calc(var(--space) / -2);
}
This works because the margin has the inverted value of the .toolbar padding.

PR #1924, opened 2026-08-26
Fix vector tiles with nested metadata could not be loaded

Phocacius:


astroidex (approved):
tested in 4.2.6. Works fine.

PR #1925, opened 2026-08-26
Improvements for new applicationswitcher element

Phocacius:
- Open link in new tab when having shift or ctrl key pressed while clicking on a link
- Replace templates when using the contextmenu => open in new tab to navigate
- Filter out invalid configurations (like references to non-existing internal applications or external links without url set), they used to throw an error 
- fallback to slug if "title" is not defined instead of throwing an error


astroidex (approved):
tested. works fine

PR #1929, opened 2026-09-01
Fix element type detection in csrf token util

Phocacius:
the result was that when multiple elements that uses csrf (MbSearchRouter, MbViewManager) were present in an application, only one of them worked.


astroidex (approved):
Tested works fine

PR #1930, opened 2026-09-01
Several batch print fixes

Phocacius:
- Fix rendering of vector tiles (application jobData property was not passed down to renderer)
- The "frames deleted" message was shown in the wrong message container
- On-page legends are rendered on every page if a "legend" container exists. Remaining legend entries are rendered grouped at the end.

### DIFF
--- a/src/Mapbender/PrintBundle/Component/PrintService.php
+++ b/src/Mapbender/PrintBundle/Component/PrintService.php
@@ -2,7 +2,6 @@
 
 namespace Mapbender\PrintBundle\Component;
 
-use setasign\Fpdi\Fpdi;
 use Mapbender\CoreBundle\Component\Application\ApplicationResolver;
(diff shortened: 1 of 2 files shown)


astroidex (changes requested):
- [ ] - NOT Tested Fix rendering of vector tiles (application jobData property was not passed down to renderer)
- [x] - OK The "frames deleted" message was shown in the wrong message container
- [x] - On-page legends are rendered on every page if a "legend" container exists. Remaining legend entries are rendered grouped at the end.

**New remarks**
1. % Überlappung seams to has an error. Try 0% and check the difference between 10% or 50%

Image

2. Entlang der Strecke Platzieren
- first print has rotation 180 (steht auf dem Kopf)

Image

Phocacius:
The frame placement will be fixed in a future PR.

PR #1933, opened 2026-09-02
 Fix sources other than WMS were not working in BaseSourceSwitcher when using shared instances

Phocacius:
e.g. when using a WMTS or VectorTiles instance that is included as a shared/free instance, they were not toggled via the BaseSourceSwitcher. That's because the logic to exctract the id from the assignemnt id was (wrongly) only applied to WMS instances, not all other sources.

This will be also merged back into the develop branch.

see internal ticket 13626


astroidex (approved):
Tested works fine

PR #1934, opened 2026-09-03
[BatchPrint] Fix automatic frame placement 

Phocacius:
The automatic placement of frames along a line was very buggy. It is now completely refactored. The new algorithm:
- start from the start of the line and try to create a bounding box (of the frame size) that includes as much as possible from the line (in contigious length), respecting a padding to the outside of the box
- for the next frame, start from where the line left the last frame (minus the configured overlap) and repeat

The automatic rotation feature is removed. This is very expensive to optimally compute. If a client really wants it, it can be reimplemented (with the new algorithm) in the future.

Also, refactored MbBatchPrintClient to make overriding of utility classes easier.


astroidex (approved):
works fine

PR #1935, opened 2026-09-03
Fix printing with custom styles

Phocacius:
Printing OGC API Features layer styled via ol-mapbox-style layers (i.e. all non-trivial styles) did not print correctly. The reason: The font size was not correctly read (in `font: normal 400 13px`, the font-weight 400 was incorrectly assumed the fontSize, resulting in way too big text), and the fill and stroke was removed it a "regular" fill/stroke was missing, however, for image geometries it must be preserved.

see internal ticket 13637


astroidex (approved):
tested. works fine

PR #1936, opened 2026-09-08
Fix simple search demo apps

Phocacius:
- Support legacy results Block in search yaml definitions
- Add alternative pin icon for search

test together with https://github.com/mapbender/mapbender-starter/pull/179

see internal ticket 13940


astroidex (approved):
Tested works fine

PR #1937, opened 2026-09-08
Fix print queue not showing after submitting print request; Default print queue directory changed to var/prints (for security reasons)

Phocacius:
(only relevant when having `mapbender.print.queueable` set to true). Print queue prints were in the default setting saved to public/prints, which resulted them to be publicly accessible via the Webserver, circumventing all access restrictions.
The new default print queue directory is application/var/prints, which is not publicly accessible. The files can still be accessed via the print queue tab, but with access control being checked. If you still want to maintain the prints publicly accessible, change the parameter `mapbender.print.queue.storage_dir` to `%kernel.project_dir%/public/prints` (or similar) in your parameters.yaml file.

Related starter PR: https://github.com/mapbender/mapbender-starter/pull/180

Also, fixed that the print queue was not showing after submitting print request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Breaking changes (for details on migration process see [UPGRADING.md]):
 * Symfony updated to version 7.4 LTS
 * Doctrine DBAL updated to version 4.4
 * Doctrine ORM updated to version 3.6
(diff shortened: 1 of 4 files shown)


astroidex (approved):
Thanks for the chenge. Very important.

PR #1938, opened 2026-09-08
[OGC API Features Layer Editor] fix column widths

Phocacius:
Add min width for max features and scale column, add soft hyphens to German translations


astroidex (approved):
Tested works fine

PR #1941, opened 2026-09-09
Move count constraint in LayersetAdminType to MapAdminType, it does not work for Overview where it's also used

Phocacius:


astroidex (approved):
tested works fine

